PR #1, opened 2017-07-20
BUG: import of psignifit did not reveal functions.

Visdoom:
Hey there Heiko,

when importing the package from a `sys.path` it only loaded the empty init file.
This should fix it now, making functions available by 
`import psignifit as ps`

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -1,6 +1,15 @@
 # -*- coding: utf-8 -*-
+#! 
 """
 Created on Sat Dec  5 20:49:14 2015
 
 @author: root
 """
+
+from .psignifit import *
+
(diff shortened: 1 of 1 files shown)


HeikoSchuett:
HI ! & welcome back!
Indeed I never tried importing this as a whole folder/package but only with the files on the path directly. Thanks!

Visdoom:
Hi!
Perfect, thanks!

HeikoSchuett:
Hey! 
I also added some code to manipulate the path as psignifit.py did not find the other submodules in some environments, when loaded with only the folder above on the path. I thiought it should work without this, but it did not.

PR #8, opened 2017-12-15
added first basic tests

HeikoSchuett:
These tests do not check whether any results are correct, but only, whether the basic function and plots run at all.


HeikoSchuett:
The plot Test does not run on travis, have to fix that first!

PR #23, opened 2020-07-19
Flask requestlayer

majorminors:
Hello,

I thought you might be interested in this.

I wanted to use psignifit with [jsPsych](https://www.jspsych.org/), so I created a simple request layer with [Flask](https://flask.palletsprojects.com/en/1.1.x/) to handle POST requests. This means I can send data from e.g. [JATOS](https://www.jatos.org/) for processing with the toolbox, and return the outputs of the toolbox to JATOS for further experimentation. Since it's in Flask, I could also just use the request layer to serve my experiment.

I thought I'd let you know in case you were interested in adding it to the repo.

There's a readme in the request_layer/ directory, and a very bare bones demo with some fake data if you wanted to check it out to see if it was suitable.

I hope it's useful.

Warm regards,

Dorian


dekuenstle:
Hello Dorian,

Thank you very much for your pull request. 
Integrating psignifit into JATOS+jsPsych as you have done is a very cool idea. 

At the moment the layer seems to be specific for your task (e.g. hard-coded `expType` and `sigmoidName`).
In order to include it in the library, we should make it more configurable and add some basic unit tests. 
I have no experience with JATOS, so I don't know how much work that would be. 
Could you imagine to generalize your requests layer? I could help with the tests.

Alternatively, you could upload the layer to a separate repository, and we could refer to it in the documentation as an advanced usage of psignifit.

What do you think?

Best
David


PS: I would be very curious about your experience with psychometric tasks on the web.

majorminors:
I actually meant to keep this quite general. I tried to make it as open as possible (i.e. you could send to this from anything that can make a POST request, not just JATOS, and you can run any library, not just jsPsych). All it needs is a JSON string of the array. Those particular hard coded bits I just took straight off the example in the wiki for psignifit python.

I'd be very happy to work with you and make it configurable. I have very little experience with unit tests, so I'd be also very happy to use this as a learning experience.

What kind of thing were you thinking for configurability?

majorminors:
Also, I'd be happy to talk about my experiences - i should warn you that it's all preparation thus far though!! I haven't actually pushed this out the door yet :(

dekuenstle:
Hey Dorian, 
Thank you, that you want to work on this! 

I think we should generalize your layer, such it can be used without writing additional python code in 80% of the cases. 
The javascript side with the JSON payload is already very general, I agree. 

Maybe it becomes clear, if we think from a user perspective: 

> The user designs an experiment with jsPsych. He uploads the experiment to JATOS or a similar service. 
> Then the user wants to evaluate his data with python-psignifit. So he/she runs the following on a server:
> ```sh
> pip install https://github.com/wichmann-lab/python-psignifit/zipball/master
> python-psignifit web-api --port=8080
> # or alternative: python3 psignifit/web_api.py --port=8080
> ```
> Afterwards, he/she can send the data from his javascript application with a POST request to myserverurl:8080 to get the psignifit result.

As said, I have no experience with JATOS. Does this make sense? 

The necessary steps are, from my perspective: 
- [ ] Basic command-line interface to configure the port. Quite easy to do this with https://docs.python.org/3/library/argparse.html
- [ ] Define the expected  JSON input (no `faked_data` field). I would recommend to stay close to psignifits expected input: `{data: [...], config: {expType: 'norm', ...}}`. Same for the return.
- [ ] Basic unit test, testing only the api. (E.g. the function names of psignifit may change in the future, we should detect this). Doable with https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch . This video shows most of the tricks: https://www.youtube.com/watch?v=ww1UsGZV8fQ .

I am happy to help with any of these!

Cheers
David

majorminors:
I see what you mean. This seems pretty straight forward. As I said, I may need help with the unit testing, but I appreciate the links. I'll see how far I get.

One concern. Flask [recommends deployment via a WGSI server](https://flask.palletsprojects.com/en/1.1.x/deploying/) for production environments. In my use case, I POST twice to the psignifit server during the course of the experiment, which then allows the experiment to continue. As such, I'd be using [gunicorn](https://gunicorn.org/) behind an nginx reverse proxy so I can run a bunch of workers in case there is more than one request incoming at a time.

Would we consider bundling a production environment option? Or just warn users out?

dekuenstle:
Good catch.
I would prefer to keep it simple for now, and link to this Flask docu site for production use.  So people can use the request layer with their preferred WGSI. 
It still should be possible to use Flask’s built-in server for basic use when one runs the layer as a script, right?

Cheers
David

majorminors:
Agreed. We can use the Flask demo to run it, and print a warning.

majorminors:
Also, quick question you may or may not be able to help me with. Security-wise, I'm limiting payloads to 1KB (which is actually probably bigger than necessary, I think), and rejecting anything other than a JSON string. We also don't evaluate the JSON string, unless I'm mistaken. I think that covers many things. But It would be nice to also verify the incoming data is the expected kind of array. I have no idea how to do this. Not even how to google it. Any tips/keywords?

dekuenstle:
I think 1KB is reasonable. This would also one of the possible (command line) parameters. 
The JSON is evaluated when you call request.get_json(), in my understanding. 

Actually, I would stick with python's "We are all responsible users": 
Expect, that the payload has the correct format and fail otherwise (as does Flask's get_json and numpy's np.asarray).
In this sense, you can also verify the data type and shape. 
I am not sure, if this is what you thought about.

Security, authentification, etc is also beyond what we can provide and is due to the user. Another good reason to reference Flask's production instructions.

majorminors:
First pass at a cli. e.g.:

`python3 request_layer.py --host 0.0.0.0 --port 8000`

Will establish a server on 0.0.0.0:8000.

I decided to just use the Flask functionality that comes with the app.run() method, rather than writing an api over the top. A little less difficult to debug if things change.

Also, this way, Flask will detect that it's being run as production and throw a warning:
```
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
```
Which I think covers our bases on that front.

What do you think?

majorminors:
So after playing around a bit, eventually my idea for the psignifit options and results was to just faithfully relay config from the user's javascript front-end and relay the result back to the front-end, again to be addressed by the user in whatever javascript code they're using. That way we don't need to maintain anything in the request_layer if psignifit changes.

However, we can't jsonify numpy arrays. I'm having trouble addressing this, probably because I have limited knowledge of numpy. As such, this might not be the right question. But I'm wondering if everything in the results structure is a numpy array, or if there's a couple of things I could deal with individually? By, e.g. extracting them, converting them, then overwriting it in the results structure.

Very open to suggestions. I've tried converting the whole thing (e.g. [this](https://www.thetopsites.net/article/51332343.shtml)) but it's proving non-trivial.

dekuenstle:
Relying on Flask's cli and warning is very good. 

Yes, the numpy arrays have to be converted before creating the result JSON. 
json.dumps supports "custom" encoder, as described in the post you linked or here https://stackoverflow.com/a/47626762 . 

Best 
David

majorminors:
Haven't gotten to the array yet.

Quick query though. Are there any other 'eval' instances in the psignifit code? Where a string from the external source might be evaluated by psignifit?

I know we said that we should abide by "python's "We are all responsible users"" ethos, but my IT department is asking, so I thought I might add validation against jsonschema to this since I have to do it anyway.

dekuenstle:
> Are there any other 'eval' instances in the psignifit code? Where a string from the external source might be evaluated by psignifit?

I am not aware of any call to the 'eval()' function or similar in python-psignifit. There is currently no JSON, YAML or PICKLE parsing. The API usually expects python objects. Simple string parsing is done to get the number of choices in a '2AFC', '3AFC', ... experiment ;-)

majorminors:
Thank you. Appreciate the response greatly!

PR #32, opened 2020-09-14
Port getStandardParameters from MATLAB implementation

dominikstrb:
The wiki for `python-psignifit` suggested implementing the getStandardParameters function: https://github.com/wichmann-lab/python-psignifit/wiki/How-to-Get-Standard-Parameters. Because this is a function I was using quite often in my own projects, I ported the MATLAB implementation (https://github.com/wichmann-lab/psignifit/blob/master/getStandardParameters.m)

### DIFF
--- a/psignifit/getStandardParameters.py
+++ b/psignifit/getStandardParameters.py
@@ -1,0 +1,73 @@
+import numpy as np
+
+import psignifit as ps
+from psignifit.utils import my_norminv, my_t1icdf
+
+
+def getStandardParameters(theta, sigmoid_name=None, widthalpha=.05):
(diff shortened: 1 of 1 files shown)


dekuenstle:
Hello Dominik,

Thanks a lot for your work. 

Please note, that we are currently working on a API cleanup, which would replace this function. 
But until then, I think your contribution is a nice utility function for everyone.

So I will merge it if there comes no objection in the next days by @HeikoSchuett , @otizonaizit or others.

dominikstrb:
Great, thanks. Looking forward to the API cleanup!

Issue #29, opened 2020-08-04
Continuous Integration


dekuenstle:
The continuous integration pipeline (currently travis) should be extended and be *mandatory* for all pull requests.
We should include:

* Unittests (pytest)
* Test coverage
* Style check (pylint or yapf)
* Demos (they should run without an error)
* Optional: Typing

PR #40, opened 2020-10-12
Continuous integration

dekuenstle:
#29


.flake8
```diff
@@ -0,0 +1,7 @@
+[flake8]
+max-line-length = 127
+max-complexity = 13
+exclude = .git,__pycache__,doc/,docs/,build/,dist/,archive/,setup.py
+ignore = E121,E123,E126,E226,E24,E704,W503,W504
```
otizonaizit, line 5: for people not using `flake8`, could you add here comments specified what kind of errors/warnings are ignored. In other words, some text explaining what every single error/warning code means (or at least a link to the corresponding flake8 URL doc) would be extremly useful

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Install package with dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -e .[tests]
+# Uncomment if sphinx docs are available in the repository
+#        pip install -e .[tests,docs]
+    - name: Lint with flake8
+      run: |
+        # exit-zero treats all errors as warnings.
+        flake8 . --count --exit-zero --statistics --show-source
```
otizonaizit, line 37: Does ``-exits-zero` mean that the `flake8` run never fails? If yes, how does the warning look like in a PR? Do I see it?

otizonaizit (approved):
Looks good. Some general comments (detailed comments are inline):
- for people coding without an IDE (most of our future contributors, I guess), it would be really useful to know what the expected style is. I don't use flake8 and don't know what the tool expects from me
- I understand you had to do it here, but in the future big commits with style-only changes should be avoided: it makes using git blame and in general understanding the history of changes quite hard
- As long as the style-checks do not make a PR fail I am fine with this and this PR can be merged
- I think it really demotivates casual contributors seeing a PR failing because their use of spacing does not conform to what `flake8` thinks it's good. I agree that a certain style consistency is definitely good, but the obnoxious style-checks typical of IDEs are in my opinion a killer for contributors less used to these tools (and we expect may of these contributors). If you are not integrating flake8 into your workflow is almost impossible to meet its style requirements, unless we relax them a lot (but maybe you are doing that already in the [`.flake8`](.flake8) file and I just do not understand it)

.flake8
dekuenstle: Did this now

.github/workflows/ci-tests.yml
dekuenstle: I made a change now, where we fail the build for "major" issues, but only print warnings to the log for style related issues. 
The log can be found at the action view. 
see the comment

dekuenstle:
The idea is, that flake8 now fails the build only for major issues, such as undefined variables, which might be missed in a unit test.
Style related issues (e.g. whitespace) are only printed to the logs.
These style related configurations shall help contributors writing nice code. 
For this, they should run flake8 themselves during development. 

We should describe this in the documentation. There we can also link to pep8 style, which is mainly what flake8 tries to assure.

Issue #35, opened 2020-10-01
Save- and loadable result object


dekuenstle:
Instead of a result dict, we want to use a result object.
It should have methods like get_threshold, get_slope, etc. (see also #33) and should store the posterior (see #34 ).
The save/load methods should handle json files and have to also (de)serialize the numpy arrays for the posterior.
In addition we have to make the configuration object serializable, such it can be added to the result object.

PR #42, opened 2020-10-20
Save and load psignifit result including configuration

dekuenstle:
#35 

@otizonaizit I just noticed, that renaming the Conf class makes the diff-view quite messy. 
Feel free to revert commit 0ce9eccf to undo this change for your review, and we can redo it afterwards.

### DIFF
--- a/psignifit/configuration.py
+++ b/psignifit/configuration.py
@@ -15,7 +15,7 @@ _PARAMETERS = {'threshold', 'width', 'lambda', 'gamma', 'eta'}
 
 
 @dataclasses.dataclass(frozen=True)
-class Conf:
+class Configuration:
     """The basic configuration object for psignifit.
 
     This class contains a set of valid options and the corresponding sanity
--- a/psignifit/configuration.py
+++ b/psignifit/configuration.py
@@ -2,7 +2,7 @@
 """
 import re
 import dataclasses
-from typing import Dict, Tuple, Optional
+from typing import Any, Dict, Tuple, Optional
 
 import numpy as np
 
@@ -66,13 +66,19 @@ class Configuration:
     uniform_weight: Optional[float] = None
 
     # parameters which are always initialized based on other parameters
-    grid_steps: Dict[str, int] = dataclasses.field(init=False)
-    steps_moving_bounds: Dict[str, int] = dataclasses.field(init=False)
+    grid_steps: Dict[str, int] = None
+    steps_moving_bounds: Dict[str, int] = None
 
     def __post_init__(self):
         self.check_parameters()
 
-    def asdict(self) -> Dict:
+    @classmethod
+    def from_dict(cls, config_dict: Dict[str, Any]):
+        config_dict = config_dict.copy()
+        return cls(confP=tuple(config_dict.pop('confP')),
+                   **config_dict)
+
+    def as_dict(self) -> Dict[str, Any]:
         return dataclasses.asdict(self)
 
     def check_parameters(self):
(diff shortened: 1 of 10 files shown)


psignifit/configuration.py
```diff
@@ -0,0 +1,216 @@
...
+        for field in dataclasses.fields(self):
+            checker_name = 'check_' + field.name
+            if hasattr(self, checker_name):
+                checker = getattr(self, checker_name)
+                checker(getattr(self, field.name))
+
+    # template for an option checking method
+    # def check_foobar(self, value):
+    #    if value > 10:
+    #       raise PsignifitException(f'Foobar must be < 10: {value} given!')
```
otizonaizit, line 94: in this template we need an example what to do when you want to set an attribute (the `object.__setattr__` dance). I am not sure the dance is still needed even if we don't use a frozen dataclass. It was needed before because we were defining a custom `__setattr__` but may not be needed anymore here

psignifit/configuration.py
```diff
@@ -0,0 +1,216 @@
...
+        # we have to set them with this special syntax
+        object.__setattr__(self, 'grid_steps', {
+            'threshold': 40,
+            'width': 40,
+            'lambda': 20,
+            # 'gamma' will be set below
+            'eta': 20,
+        })
+        if value == ExperimentType.YES_NO.value:
+            self.grid_steps['gamma'] = 20
```
otizonaizit, line 158: see my first comment above about using a frozen dataclass

psignifit/result.py
```diff
@@ -0,0 +1,52 @@
...
+@dataclasses.dataclass(frozen=True)
+class Result:
+    sigmoid_parameters: Dict[str, float]
+    configuration: Configuration
+
+    @classmethod
+    def from_dict(cls, result_dict: Dict[str, Any]):
+        result_dict = result_dict.copy()
+        config_dict = result_dict.pop('configuration')
+        # Numpy arrays should be initialized from lists here.
```
otizonaizit, line 27: is this comment a relic from the past or is some code missing here?

otizonaizit (approved):
This is quite cool. I have some philosophical comments below, but otherwise it's good to merge!

psignifit/configuration.py
```diff
@@ -0,0 +1,216 @@
...
+
+from . import sigmoids
+from .utils import PsignifitException
+from .typing import ExperimentType, Prior
+
+
+_PARAMETERS = {'threshold', 'width', 'lambda', 'gamma', 'eta'}
+
+
+@dataclasses.dataclass(frozen=True)
```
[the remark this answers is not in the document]
dekuenstle: Yes, the intention was to avoid users shooting themselves in the foot. 
I also see the concerns you state, and will remove the frozen.

psignifit/configuration.py
dekuenstle: As I 'unfreezed' the datclassees, we dont have to change anything here.

psignifit/configuration.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def as_dict(self) -> Dict[str, Any]:
+        return dataclasses.asdict(self)
+
+    def check_parameters(self):
+        for field in dataclasses.fields(self):
+            checker_name = 'check_' + field.name
+            if hasattr(self, checker_name):
+                checker = getattr(self, checker_name)
+                checker(getattr(self, field.name))
```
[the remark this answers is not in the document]
dekuenstle: I made it clear with a docstring and expressive variable names.

psignifit/result.py
dekuenstle: I made this more clear.

Issue #33, opened 2020-10-01
Sigmoid classes


dekuenstle:
Sigmoid classes should have functions to calculate the inverse, slope. Tests for these functions have to be added. 
This was started in #31.

Refactoring these functions:
https://github.com/wichmann-lab/python-psignifit/blob/7b5bf14fa3edd770be9e05966843b05bdd91ec6e/psignifit/psignifit.py#L417

PR #43, opened 2020-11-03
Sigmoids

dekuenstle:
#33


psignifit/demos/plot_all_sigmoids.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+STIMULUS_LEVELS = np.linspace(1e-12, 1-1e-12, num=10000)
+M = 0.5
+WIDTH = 0.9
+PC = 0.5
+ALPHA = 0.05
+
+
+def plot_sigmoid(sigmoid, name, x, threshold=M, width=WIDTH, axes=None):
```
otizonaizit, line 14: very cool!

psignifit/sigmoids.py
```diff
@@ -3,16 +3,20 @@
 If you add a new sigmoid type, add it to the CLASS_BY_NAME constant
 and to the _LOGSPACE_NAMES, if it expects stimulus on an exponential scale.
 """
+from typing import Optional, TypeVar
+
 import numpy as np
-from typing import Optional
+import scipy
```
otizonaizit, line 9: by convention this is typically imported like
```python
import scipy as sp`
```
also, `scipy` is a lazy-import package, so given that below you are using `scipy.stats` you'll need to import it here explicitly:
```python
import scipy as sp
import sp.stats
```
The tests are passing probably because `scipy.stats` is imported somewhere else in the code, but not in this file.

psignifit/tests/test_sigmoids.py
```diff
@@ -42,26 +42,47 @@ def test_sigmoid_by_name(sigmoid_name):
 
 @pytest.mark.parametrize('sigmoid_name', sigmoids.ALL_SIGMOID_NAMES)
 def test_sigmoid_sanity_check(sigmoid_name):
```
otizonaizit, line 44: very cool! I think this is enough basic testing also for future sigmoids. Maybe we could expose this to the user so that they can test "home-made" sigmoids? Or would this be too much of a mess?

otizonaizit (changes requested):
This is a very clean rewrite. I am a bit worried about future maintenance for scientists not used to OOP, but the result is so much cleaner than the previous code that I am willing to take the risk ;-)

psignifit/tests/test_sigmoids.py
dekuenstle: Great idea! Did this

Issue #34, opened 2020-10-01
Refactor confidence interval calculation


dekuenstle:
This requires to store the posterior also in the result object.

https://github.com/wichmann-lab/python-psignifit/blob/refactor-api/psignifit/getConfRegion.py

PR #44, opened 2020-11-04
Confidence

dekuenstle:
#34 

I do not consider this as finished, but  it would be already great to get some feedback on this.
Especially about how I calculate the CI now, it would be great if you could have a second look if it is correct
and if you see a better way to test it.


otizonaizit:
I'll need a bit more time to review this. If you do other work in this direction keep on commiting here.

psignifit/psignifit.py
```diff
@@ -178,11 +186,8 @@ def _fit_parameters(data: np.ndarray, bounds: ParameterBounds,
     # do first sparse grid posterior_grid evaluation
     grid = get_grid(bounds, steps_moving_bounds)
     posteriors_sparse, grid_max = posterior_grid(data, sigmoid=sigmoid, priors=priors, grid=grid)
-    # normalize the posterior_grid
-    posterior_volumes = posteriors_sparse * integral_weights([grid_value for _, grid_value in sorted(grid.items())])
-    posterior_integral = posterior_volumes.sum()
     # indices on the grid of the volumes that contribute more than `tol` to the overall integral
-    mask = np.nonzero(posterior_volumes / posterior_integral >= max_bound_value)
+    mask = np.nonzero(posteriors_sparse >= max_bound_value)
```
otizonaizit, line 190: This needs to be refactored.

  - document what `max_bound_value` is
  - factor the loop out to a different function called `tighten_bounds`, which should be documented properly

psignifit/getConfRegion.py
```diff
@@ -48,12 +48,111 @@
...
 # }
+from typing import Sequence
 
 import numpy as np
+from scipy import stats
 
 from .marginalize import marginalize
 
 
+def confidence_intervals(probability_mass: np.ndarray, grid_values: np.ndarray,
```
otizonaizit, line 59: `probability_mass` is now hidden within the `_fit_parameters` function. How do we get it here?

  - we are actually calculating the confidence intervals already within the function `psygnifit`, where we have access to the `probability_mass`
  - should we attach `probability_mass` to `results`? But then it is a huge array that we are carrying arround
  - or, make the storing of the `probability_mass` optional, so that you can only calculate the confidence intervals for different parameters if you have optionally stored the `probability_mass`

psignifit/psignifit.py
```diff
@@ -125,7 +131,9 @@ def psignifit(data, conf=None, **kwargs):
     # plot.plotPsych(result)
 
     return Result(sigmoid_parameters=fit_dict,
-                  configuration=conf)
+                  configuration=conf,
+                  confidence_intervals=intervals_dict,
+                  posterior=posteriors)
```
otizonaizit, line 136: look into `getConfRegion.py` for a discussion about `posteriors`

psignifit/getConfRegion.py
```diff
@@ -48,12 +48,111 @@
...
+    Args:
+        probability_mass: Probability mass at each grid point, shape (n_points, n_points, ...)
+        grid_values: Parameter values along grid axis, shape (n_dims, n_points)
+        credible_mass: Minimal mass in highest density region
+    Returns:
+        Grid value at interval per dimension, shape (n_dims, 2)
+
+    .. _stats.stackexchange: https://stats.stackexchange.com/questions/148439/what-is-a-highest-density-region-hdr
+    """
+    decreasing_mass = np.sort(np.ravel(probability_mass))[::-1]
```
otizonaizit, line 120: change the use of numpy functions to using array methods everytime this is possible. For example:
`np.ravel(a)` ➔ `a.ravel()`

otizonaizit, line 120: Use `a.reshape(-1)` instead of `a.ravel()` everytime we need a view (`ravel` copies the underlying data). Look at the `np.ravel` documentation for more info:

> When a view is desired in as many cases as possible, `arr.reshape(-1)` may be preferable.

psignifit/getConfRegion.py
```diff
@@ -48,12 +48,111 @@
...
+
+def percentile_intervals(probability_mass: np.ndarray, grid_values: np.ndarray, p_value: float):
+    """ Percentile intervals on a grid.
+
+    Find alpha/2 and 1-alpha/2 percentiles on marginal probability mass (alpha = 1 - p_value).
+    
+    Args:
+        probability_mass: Probability mass at each grid point, shape (n_points, n_points, ...)
+        grid_values: Parameter values along grid axis, shape (n_dims, n_points)
+        credible_mass: Minimal mass in highest density region
```
otizonaizit, line 144: this parameter is not used here!

psignifit/likelihood.py
```diff
@@ -5,7 +5,7 @@
 import scipy.special as sp
 from scipy import optimize
 
-from .utils import fp_error_handler, PsignifitException
+from .utils import fp_error_handler, PsignifitException, integral_weights
```
otizonaizit, line 8: the function `integral_weights` should be moved to the `likelihood.py` file, if nothing else uses it.

psignifit/likelihood.py
```diff
@@ -41,7 +41,12 @@ def posterior_grid(data, sigmoid: Sigmoid, priors: Dict[str, Prior],
             grid_max[name] = None
         else:
             grid_max[name] = grid_values[index]
-    return p, grid_max
+
+    # normalize the posterior_grid
+    posterior_volumes = p * integral_weights([grid_value for _, grid_value in sorted(grid.items())])
+    posterior_integral = posterior_volumes.sum()
```
otizonaizit, line 47: Is this sum safe? Are we summing too many little numbers with some big numbers and so losing precision?

psignifit/getConfRegion.py
```diff
@@ -48,12 +48,111 @@
...
+    if not np.isclose(probability_mass.sum(), 1):
+        raise ValueError(f'Expects sum(probability_mass) to be 1., got {probability_mass.sum():.4f}')
+    intervals = np.empty((probability_mass.ndim, len(p_values), 2))
+    for p_ix, p_value in enumerate(p_values):
+        intervals[:, p_ix, :] = calc_ci(probability_mass, grid_values, p_value)
+
+    return intervals
+
+
+def grid_hdi(probability_mass: np.ndarray, grid_values: np.ndarray, credible_mass: float) -> np.ndarray:
```
otizonaizit, line 100: this stuff can be easily tested in 1-D and 2-D by using a Gaussian probability mass, where we know the expected confidence intervals

psignifit/psignifit.py
dekuenstle: Once working on this, I preferred another solution: I called the new variable thight_bounds, but the function mask_bounds.
The function is relatively general, has nothing to do with tight or loose, just setting bounds around a mask.

dekuenstle:
Hey @otizonaizit ,

it took me some more weeks than expected until I found time to implement your suggestions.
Could you please have another look, and either accept or comment? Thank you!

Cheers 
David

HeikoSchuett:
Just a small comment on keeping the posterior grid around: 
I do keep that in the MATLAB version for convenience and once you give the user the option to remove it after running the main psignifit inference that should not be a terrible performance problem. Also there are a few detailed analyses like looking at the joint marginals between parameters for example that require the full joint posterior. Thus, I would not necessarily delete the joint posterior by default even though it is large. Just give the user the option to delete it easily, if they run into memory issues.
For the credible interval calculations in this pull request this would not be extremely important though: The default CIs, which we tested in the paper, are calculated based only on the marginals. Thus, it would not be a big problem to remove the joint posterior before this calculation. Calculating the credible intervals by projecting down the higher dimensional confidence regions is unorthodox enough that we don't have to support it.

PR #57, opened 2021-07-05
small bugfix to estimateType == 'mean'

emilaz:
There's a small bug when trying to use estimateType 'mean'. Might be translation error from matlab.


dekuenstle:
Thank you @emilaz !

PR #59, opened 2021-08-30
getSlopePC parenthesis location fix

ddeloss:
getSlopePC has a parenthesis in the wrong place


ddeloss:
Please take a look and merge in if all looks well. This doesn't cause any issues when looking at the slope at the midpoint because `np.log(1/0.5 - 1)` is 0. So the issue only pops up when actually specifying a percent correct that isn't at the midpoint.

HeikoSchuett (approved):
you are right, thanks for finding this.

PR #66, opened 2022-01-13
Matlab test

dekuenstle:


otizonaizit:
Impressive work, @dekuenstle : Kudos!

psignifit/tests/test_matlab.py
```diff
@@ -0,0 +1,77 @@
...
+    if options['expType'] == 'equalAsymptote':
+        # matlab returns gamma filled with Nones, but python expects gamma==lambda
+        mat_ci[3] = mat_ci[2]
+    # matlab dims are (n_param, 2, n_ci), but python expects (n_param, n_ci, 2),
+    mat_ci = np.transpose(mat_ci, (0, 2, 1))
+    py_ci = param_pydict2matlist(pyresults.confidence_intervals)
+    np.testing.assert_allclose(py_ci, mat_ci, atol=0.001, rtol=2)
+
+
+@pytest.mark.skip(reason="Testing all cases takes approx 30min, should be run explicitly.")
```
pberkes, line 66: Marking this tests as `skip` means that they are not run, not even in the continuous integration workflow. Should we add a pytest marker like `slow`, or a customer one like `matlab`, so that we can run these tests selectively from the command line?

pberkes (commented):
Pretty cool, a suggestion about skipping the tests

psignifit/_parameter.py
```diff
@@ -77,6 +77,10 @@ def masked_parameter_bounds(grid: Dict[str, Optional[np.ndarray]], mesh_mask: np
     new_bounds = dict()
     mask_indices = mesh_mask.nonzero()
     for axis, (parameter_name, parameter_values) in enumerate(sorted(grid.items())):
+        if parameter_values is None:
+            new_bounds[parameter_name] = None
+            continue
```
pberkes, line 82: are these checks for the benefit of the Matlab tests alone? in this case, we need a different solution

(the same is true for all other similar checks)

psignifit/_result.py
```diff
@@ -18,7 +18,7 @@ def default(self, obj):
 
 @dataclasses.dataclass
 class Result:
-    parameter_estimate: Dict[str, float]
+    parameter_fit: Dict[str, float]
```
pberkes, line 21: what's the idea behind this name change?

psignifit/psignifit.py
```diff
@@ -83,31 +83,36 @@ def psignifit(data: np.ndarray, conf: Optional[Configuration] = None,
...
 
-    grid_values = [grid_value for _, grid_value in sorted(grid.items())]
-    intervals = confidence_intervals(posteriors, grid_values, conf.confP, conf.CI_method)
-    intervals_dict = {param: interval_per_p.tolist()
-                      for param, interval_per_p in zip(sorted(grid.keys()), intervals)}
+    grid_none_ix = tuple(ix for ix, (param, value) in enumerate(sorted(grid.items())) if value is None)
+    grid_params = [param for param, value in grid.items() if value is not None]
+    grid_values = [grid[param] for param in grid_params]
+    intervals = confidence_intervals(np.squeeze(posteriors, axis=grid_none_ix), grid_values, conf.confidence_intervals, conf.CI_method)
+    intervals_dict = {param: intervals[ix].tolist() for ix, param in enumerate(grid_params)}
```
pberkes, line 98: is this needed, or for Matlab compatibility?

psignifit/psignifit.py
```diff
@@ -83,31 +83,36 @@ def psignifit(data: np.ndarray, conf: Optional[Configuration] = None,
...
     marginals = marginalize_posterior(grid, posteriors)
 
     if conf.verbose:
         _warn_marginal_sanity_checks(marginals)
 
+    if fit_dict['gamma'] is None:  # equal asymptotes
+        fit_dict['gamma'] = fit_dict['lambda']
+        intervals_dict['gamma'] = intervals_dict['lambda']
```
pberkes, line 106: we checked, and confirm this is correct

pberkes (changes requested):
Reviewed with @otizonaizit , let's discuss the open questions

psignifit/tests/test_matlab.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # INTERVAL ESTIMATE
+    mat_ci = results['conf_Intervals']
+    if options['expType'] == 'equalAsymptote':
+        # matlab returns gamma filled with Nones, but python expects gamma==lambda
+        mat_ci[3] = mat_ci[2]
+    # matlab dims are (n_param, 2, n_ci), but python expects (n_param, n_ci, 2),
+    mat_ci = np.transpose(mat_ci, (0, 2, 1))
+    py_ci = param_pydict2matlist(pyresults.confidence_intervals)
+    np.testing.assert_allclose(py_ci, mat_ci, atol=0.001, rtol=2)
```
pberkes, line 63: rtol = 0.1

psignifit/tests/test_matlab.py
```diff
@@ -0,0 +1,77 @@
...
+    #     rtol=1 assures the same order of magnitude
+    #
+    # These matlab results are currently not tested:
+    # 'X1D', 'logPmax', 'integral', 'marginals', 'marginalsX', 'marginalsW', 'data',
+    # 'devianceResiduals', 'deviance', 'Cov', 'Cor', 'timestamp', 'meanFit', 'MAP'
+
+    ## POINT ESTIMATE
+    np.testing.assert_allclose(param_pydict2matlist(pyresults.parameter_fit),
+                               results['Fit'],
+                               atol=0.001, rtol=1)
```
pberkes, line 53: rtol = 0.05

psignifit/_configuration.py
```diff
@@ -64,10 +63,15 @@ class Configuration:
     def __post_init__(self):
         self.check_attributes()
 
+    @classmethod
+    def from_matlab_options(cls, option_dict: Dict[str, Any], **kwargs):
+        python_configs = config_from_matlab(option_dict, **kwargs)
+        return cls(**python_configs)
```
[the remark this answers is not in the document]
dekuenstle: I removed this class function but exposed `psignifit.config_from_matlab`. This might be handy for users switching from MATLAB to Python

psignifit/_matlab.py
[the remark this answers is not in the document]
dekuenstle: See above -> Matlab compatiblitiy can be useful for people switching from matlab

psignifit/_parameter.py
dekuenstle: No, they handle special cases where no bounds are provided.

psignifit/_result.py
dekuenstle: Easier to recognize the content for switcher

psignifit/psignifit.py
dekuenstle: Yes, this is needed. Even non-matlab users might not provide all bounds. This case handels these

dekuenstle: I removed this

psignifit/_parameter.py
dekuenstle: Now after a refactor of fixed_parameters, bounds should never be none. Removed this.

psignifit/tests/test_matlab.py
dekuenstle: done

dekuenstle:
- [ ] Configuration should raise Exception if sigmoid name starts with log (notice user, that we do not support logspace, transform on their own)
- [ ] Only run these tests on non-logspace names. 
- [ ] Investigate the difference between MATLAB and Python with this datafile. With the demo-file, we get exactly the same results.

dekuenstle:
I dug deeper into the difference between the MATLAB and psignifit results on these datasets. 

1. In the test I forgot to pool the data blocks. After introducing the pooling the point estimates can typically recovered with <1% error.
2. I changed the default CI method to `percentiles` (see #133 ) to obtain almost the same width (<10% difference). 
3. The python CI still remain systematically smaller (~5%). In my perspective this is not a bug: While the MATLAB sets the border in the middle of two grid points (one within, one outside the CI), while the Python version interpolates the cumulative distribution. In steep regimes like the one below, the interpolation must be more narrow.

psignifit/_priors.py
```diff
@@ -170,8 +170,6 @@ def setup_priors(custom_priors, bounds, stimulus_range, width_min, width_alpha,
         priors.update(custom_priors)
     check_priors(priors, stimulus_range, width_min)
 
-    for parameter, prior in priors.items():
-        priors[parameter] = normalize_prior(prior, bounds[parameter])
+    for parameter, bound in bounds.items():
+        priors[parameter] = normalize_prior(priors[parameter], bound)
     return priors
```
otizonaizit, line 175: I am not sure I understand why this was changed...

psignifit/psignifit.py
```diff
@@ -81,22 +81,25 @@ def psignifit(data: np.ndarray, conf: Optional[Configuration] = None,
...
     intervals_dict = {param: interval_per_p.tolist()
                       for param, interval_per_p in zip(sorted(grid.keys()), intervals)}
     marginals = marginalize_posterior(grid, posteriors)
 
     if conf.verbose:
         _warn_marginal_sanity_checks(marginals)
 
+    if 'gamma' not in fit_dict:  # equal asymptotes
+        fit_dict['gamma'] = fit_dict['lambda']
+        intervals_dict['gamma'] = intervals_dict['lambda']
```
otizonaizit, line 101: this is taken care of, at least partially in the following lines (106-110), so line 101
`intervals_dict['gamma'] = intervals_dict['lambda']` should be added after line 110 below

dekuenstle: Actually, given its age this PR is too full and does too many things. I currently see it more as a playground to discover if/why there are differenes between the MATLAB and Python version. Afterwards IMHO the essence of this PR should be reimplemented in a separate branch.

psignifit/_priors.py
dekuenstle: I assume this is a relict from a time when bounds might not be defined for fixed items. Therefore, the former version would crash. If I'm right, you changed the behavior last week such that fixed parameters fill bounds?!

otizonaizit:
Obsoleted by #233

Issue #67, opened 2023-01-02
Error in core function 


tomwallis:
Hi team,

I could not find info on which versions of Python are supported by `psignifit`. Tests fail for me with the following under Python 3.8, 3.9 and 3.10:

```python
============================= test session starts ==============================
platform darwin -- Python 3.9.15, pytest-7.2.0, pluggy-1.0.0
rootdir: XXX
plugins: anyio-3.6.2
collected 1 item / 1 error

==================================== ERRORS ====================================
_ ERROR collecting miniforge3/envs/prior_object_knowledge/lib/python3.9/site-packages/psignifit/tests/test_basic.py _
../../../../../miniforge3/envs/prior_object_knowledge/lib/python3.9/site-packages/numpy/core/fromnumeric.py:1565: in squeeze
    squeeze = a.squeeze
E   AttributeError: 'list' object has no attribute 'squeeze'

During handling of the above exception, another exception occurred:
../../../../../miniforge3/envs/prior_object_knowledge/lib/python3.9/site-packages/psignifit/tests/test_basic.py:28: in <module>
    res=ps.psignifit(data,options)
../../../../../miniforge3/envs/prior_object_knowledge/lib/python3.9/site-packages/psignifit/psignifit.py:265: in psignifit
    result = psignifitCore(data,options)
../../../../../miniforge3/envs/prior_object_knowledge/lib/python3.9/site-packages/psignifit/psignifit.py:384: in psignifitCore
    result['marginals'] = np.squeeze(result['marginals'])
<__array_function__ internals>:200: in squeeze
    ???
../../../../../miniforge3/envs/prior_object_knowledge/lib/python3.9/site-packages/numpy/core/fromnumeric.py:1567: in squeeze
    return _wrapit(a, 'squeeze', axis=axis)
../../../../../miniforge3/envs/prior_object_knowledge/lib/python3.9/site-packages/numpy/core/fromnumeric.py:43: in _wrapit
    result = getattr(asarray(obj), method)(*args, **kwds)
...
=========================== short test summary info ============================
ERROR ../../../../../miniforge3/envs/prior_object_knowledge/lib/python3.9/site-packages/psignifit/tests/test_basic.py - ValueError: setting an array element with a sequence. The requested array h...
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
========================= 6 warnings, 1 error in 0.17s =========================
```

[...]

PR #68, opened 2023-01-09
Fix bug on squeezing marginal distributions

tomwallis:
This change appears to fix [issue #67](https://github.com/wichmann-lab/python-psignifit/issues/67).

Core functions were returning an error in computing marginal distributions because `np.squeeze` was being called on a list. It's strange that this bug shows up now where presumably many other users should have encountered it. I suspect it might be caused by some update to `np.squeeze` to change its behaviour.

I have now adjusted to call on the `np.array` output of marginalize before entering in the list. Psignifit tests now pass, and `demo_001.py` produces a plot. 

However, I have not extensively tested whether this preserves the intended functionality of `marginalize` and the `results["marginals"]` values, nor whether this is specific to different versions of Python / numpy. Advice welcome.

### DIFF
--- a/psignifit/psignifit.py
+++ b/psignifit/psignifit.py
@@ -377,13 +377,9 @@ def psignifitCore(data, options):
     
     for idx in range(0,d):
         m, mX, mW = marginalize(result,np.array(idx))
-        result['marginals'].append(m)
-        result['marginalsX'].append(mX)
-        result['marginalsW'].append(mW) 
-    
-    result['marginals'] = np.squeeze(result['marginals'])
-    result['marginalsX'] = np.squeeze(result['marginalsX'])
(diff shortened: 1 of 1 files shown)


HeikoSchuett:
This seems like a fine fix to me. I would not have thought that this breaks based on newer versions, but if it fixes the bug that's great!

Thank you for finding and fixing this!

dekuenstle:
Hi @tomwallis 
thanks for posting the bug; I agree with your fix. 
`np.squeeze`, indeed, did change its behaviour and now just works on lists with equal length entries.  For example, `np.squeeze([[1], [2],  [3]])` works but `np.squeeze([[1], [2, 3]])` fails intentionally.

tomwallis:
Ok, thanks for checking! I can't merge, so I'll leave it to you.

PR #69, opened 2023-01-16
Add python CI tests

tomwallis:
Set up Github workflow to build and run tests on different Python versions. Should be set to trigger with push to master or pull request to master. 

I also removed the `travis.yml` configuration, which appeared to be outdated. 

Note that the URL in README should be updated to point to the `wichmann-lab` repo, not to this one.


tomwallis:
Perhaps this is not required, but I didn't understand why your checks often didn't complete (see e.g. previous older pull requests) or appear not to be run by pull requests to master. I am not an expert on setting up Github actions, but perhaps this helps to bring CI up to speed here.

PR #72, opened 2024-06-21
Empty plot fix

lschwetlick:
Whenever the library was imported in a notebook, it would cause an empty plot to appear. The reason was the use of plt.gca() in the function definition line (in the default arguments). This fix moves the call into the function such that it is not executed during import.


pberkes (approved):
LGTM

Issue #79, opened 2024-09-12
Test for marginal bounds is not meaningful for some parameters


pberkes:
The library makes some after-the-fact checks to warn the user about potential problems with the calculations.

In the function `_warn_marginal_sanity_checks`, it is checked that the posterior over each parameter is at its peak 1000 times larger than at the limit of its domain.

This is the code:
```
        peak = marginal.max() / 1000
        if marginal[0] > peak or marginal[-1] > peak:
            warnings.warn(f'''psignifit:boundWarning\n
```

However, the maximum posterior of a parameter might well be at the extremes of that interval. For instance, if the data is not overdispersed, the posterior of the parameter eta should peak at 0, its lowest bound.

Simulation on non-overdispersed data therefore always produce the following unjustified warning:

```
/Users/pietro.berkes/o/psignifit/psignifit/psignifit.py:190: UserWarning: psignifit:boundWarning

                             The marginal on the eta bounds is not smaller than 1/1000 of the peak
                             it means that the prior of the eta parameter has an influence to the result.

                             This means (1) the prior may be too narrow, or (2) you know what you are doing.

                             If you were using the default settings, this is a bug in the software or your data
                             are highly unusual,
                             if you changed the defaults please check the priors and parameter bounds.
```

How to reproduce

```
n_trials = 20
data= np.array([
    [-0.18334863789409236,  0., n_trials],
    [  0.2899908172635446,  0., n_trials],
    [  0.7633302724211816,  5., n_trials],
    [  1.2366697275788185, 15., n_trials],
    [  1.7100091827364556, 20., n_trials],
    [  2.1833486378940927, 20., n_trials],
    ]
)

conf = ps.Configuration(sigmoid='logistic', experiment_type=ps._typing.ExperimentType.YES_NO.value)
res = ps.psignifit(datal, conf)
```

pberkes:
The same might be true for `gamma` and `lambda`, to be checked. We need to clarify what the intention of the test is to understand how to fix the code.

pberkes:
Felix wrote:
"""
The MATLAB version only does the border check for thresholds and slopes -- not for all five parameters, because, as you wrote, the others are often "legally" highest at the boundary.

A very simple solution -- remove the superfluous warnings for lamda, gamma and eta.
"""

pberkes:
For reference, this is the equivalent MATLAB code:
```
%% after processing

%check that the marginals go to nearly 0 at the borders of the grid
if options.verbose > -5
[...]

PR #99, opened 2024-09-19
Removes unnecessary warnings on marginal checks

guillermoaguilar:
for the parameters other than threshold and width this check does not make sense

Issue #79


pberkes (approved):
LGTM

PR #110, opened 2024-09-20
Fix most warnings

pberkes:
Now reduced to 9 warnings when running tests (down from several hundreds)

WARNING!!! the change in `psignifit` has got to be reviewwd


psignifit/psigniplot.py
```diff
@@ -46,7 +46,7 @@ def plot_psychometric_function(result: Result,  # noqa: C901, this function is t
     if plot_data:
         y_data = data[:, 1] / data[:, 2]
         size = np.sqrt(data_size / 2 * data[:, 2])
-        ax.scatter(x_data, y_data, s=size, c=data_color, marker='.', clip_on=False)
+        ax.scatter(x_data, y_data, s=size, color=data_color, marker='.', clip_on=False)
```
dekuenstle, line 49: This might fail if data_color is a list/array of colors

dekuenstle (changes requested):

psignifit/psigniplot.py
pberkes: we decided to only support one color; I changed the type of `data_color`

PR #116, opened 2024-09-20
Fixed params

lschwetlick:
- Add the ability to also fit threshold and width. 
- fixes the tests to check the fitting with fixed parameters
- add compatibility for numpy <2


pberkes (approved):
Brilliant!

PR #132, opened 2024-09-21
Really silence warnings in tests

otizonaizit:
Removed the last sources of spurious warnings emitted while running tests.

@lschwetlick : please could you double check this commit ec4e71629b9974f90343a2fcd7ae8ab0046681ce ?
That warning was added by you but I actually think it is wrong. The warning was issued when in an experiment type different from equal asymptote the lambda and gamma where both set but different? If at all it should be the other way around, in the sense that we should warn _only_ in the equal asymptote case. Or am I misunderstanding something?


lschwetlick:
You are absolutely right, I had an error in the logic there! LGTM

lschwetlick (approved):

Issue #91, opened 2024-09-19
Fix plots functions


guillermoaguilar:
- [X] plot psychometric function by itself
- [X] model fit plot with three panels, psy. fx, shape check and time dependence 
- [x] 1d marginal, for all parameters
- [x] priors plot, 3 panels for threshold, width and lambda
- [x] 2D marginal plot
- [x] plot Bayes

guillermoaguilar:
Closed by #138

PR #138, opened 2024-09-23
fixes all plot functions

guillermoaguilar:
-  2D marginal plot. Automatic swapping of axes is removed because it was a very confusing behavior and only depended on the alphabetical order of the parameters (indices).
- priors plot fixed with same range as in MATLAB version. Previous version to set the range was hacky.
- bayes plot, which was missing in the python implementation, now implemented for all combinations of non-fixed parameters.


guillermoaguilar:
This will close Issue #91

PR #139, opened 2024-09-24
fixes threshold method, order of variables gamma and lambda was reversed

guillermoaguilar:


pberkes (changes requested):
The parameters were clearly swapped, so this is a fix.

I'm concerned about the fact this:
- the test was passing before, how was that possible?
- the new test uses `unscaled=True`, which sets gamma and lambda to 0, and therefore does not test the fix

pberkes:
BTW, the new test also passes with the old version of the code. We need to have a test that fails with the old version

pberkes:
it's because `result.slope_at_proportion_correct` does the same as we manually do in the test, so we are not actually testing anything

guillermoaguilar:
yeah I just fixed the test by adding unscale=True, but I don't understand what was the logic of the test in the first place....

guillermoaguilar:
asking for help because I'm lost with that test

pberkes:
I can have a look later, if it failed after it was fixed, it might mean that there is a similar error somewhere else

pberkes (approved):

pberkes:
Hi Guillermo, I improved the previous test, and added one test that would have failed with gamma and lambda switched. See you like it

guillermoaguilar (commented):
it looks good to me now. The test makes sense now

PR #144, opened 2024-09-24
Change the infrastucture of the docs to jupyterbook

lschwetlick:
I am starting to fix the user guides section 
- [x] Experiment Types
- [x] index
- [ ] interval bias
- [ ] options-dictionary
- [ ] plot-functions
- [ ] pooling-utility
- [ ] priors
- [ ] result-struct


lschwetlick:
I might need some help- I left a bunch of TODOs in the options-dictionary file, because I didnt know enough about them. Especially the significance of the moving bounds...

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,32 @@
...
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Setup Miniconda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          auto-update-conda: true
+          auto-activate-base: false
+          miniconda-version: 'latest'
```
otizonaizit, line 21: why are we installing `conda` at all? Further down we use `pip` to install things...
Is this a relic of some other project or is `conda` really needed for something?

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,32 @@
...
+          python-version: 3.11
```
otizonaizit, line 22: python 3.13 has been released recently. I think it's better to build the docs with the latest version of Python.

otizonaizit:
@guillermoaguilar : thanks!!! It's exciting to see this thing take form :-)

.github/workflows/build-docs.yml
guillermoaguilar: you're right, this was carried over from another project. Removing...

.github/workflows/build-docs.yml
guillermoaguilar: changed.

Issue #141, opened 2024-09-24
The implementation of the negative distributions is wrong


pberkes:
For example, the value of a negative sigmoid at x, is not `1 - sigmoid(x)`, as in the current code (see below), but `sigmoid(1 - x)` (or else I'm misunderstanding what "negative" is).

In the case of symmetrical sigmoids with threshold = 0.5, they are the same, which is why the tests were passing.

```
    def __call__(self, stimulus_level: N, threshold: N, width: N) -> N:
        """ Calculate the sigmoid value at specified stimulus levels. """
        value = self._value(stimulus_level, threshold, width)

        if self.negative:
            return 1 - value
        else:
            return value
```

pberkes:
I was wrong, negative seems to mean `1 - sigmoid(x)`. Closing

pberkes:
Reopening: one of the reasons I was confused is that in `Sigmoid.__init__` we have 

```
if negative:
    self.PC = 1 - PC
else:
    self.PC = PC
```

which suggests that negative sigmoids are switched along the x-axis (sigmoid(1 - x)) instead of the y-axis (1 - sigmoid(x)).

The current code would make the threshold for the sigmoid below at PC=0.6 to be something like 3, while I expect 6
![telegram-cloud-photo-size-2-5447644065236640566-y](https://github.com/user-attachments/assets/3fcc94e6-569d-4c16-b0d8-b96bc2c3305b)

I think the code is incorrect, and will submit tests to prove it ;-)

otizonaizit:
I think with this kind of things it is important to double check what the MATLAB version is doing.
There are two possible outcomes:

1. MATLAB behaves differently: this indicates that the problem is in the porting, in which case we can just fix things without further ado, or
2. MATLAB is doing the same thing: we should ask @FelixWichmann and @HeikoSchuett if the observed behavior is a bug or a feature, and only then proceed with fixing things

The same argument holds for  #143, #145, #137, #109, and probably others.

In this particular case "negative" sigmoid is probably a field-specific term, so I think it's safer to just ask @FelixWichmann and @HeikoSchuett instead of guessing what it may mean...

otizonaizit:
Of course acting as I describe above is a lot of effort, especially because the APIs are so different that it is not necessarily obvious how to replicate things in MATLAB. On the other hand, I don't see an easier way to approach this issue.

guillermoaguilar:
I don't think the python implementation is wrong actually.

[...]

PR #146, opened 2024-09-25
Improve sigmoid tests, fix bug in Gaussian._slope

pberkes:
[updated]

Things that happened in this PR:
- I added tests for all the existing sigmoids that make sure that their implementation is correct. This includes _value, _slope, and _inverse
- I made the sigmoid tests more generic, they were testing always with PC=0.5, alpha=0.05, and threshold=0.5 . These settings do not exercise the code outside of the default values, which is why we missed a problem with negative sigmoids
- I refactored test code (`assert_sanity_checks`) from the class Sigmoid to a standalone function
- The value of Sigmoid.PC now remains consistent even for negative sigmoids, and can be used for plotting and other computations. The _value, _slope, and _inverse method should use Sigmoid._PC, which for negative sigmoid is equal to `1 - PC`

### DIFF
--- a/psignifit/sigmoids.py
+++ b/psignifit/sigmoids.py
@@ -147,8 +147,8 @@ class Sigmoid:
 class Gaussian(Sigmoid):
     """ Sigmoid based on the Gaussian distribution's CDF. """
     def _value(self, stimulus_level, threshold, width):
-        C = width / (norminv(1 - self.alpha) - norminv(self.alpha))
-        return normcdf(stimulus_level, (threshold - norminvg(self._PC, 0, C)), C)
+        C =(norminv(1 - self.alpha) - norminv(self.alpha))
+        return normcdf(stimulus_level, (threshold - norminvg(self._PC, 0,  width / C)),  width / C)
 
     def _slope(self, stimulus_level: np.ndarray, threshold: np.ndarray, width: np.ndarray) -> np.ndarray:
         C = norminv(1 - self.alpha) - norminv(self.alpha)
--- a/psignifit/sigmoids.py
+++ b/psignifit/sigmoids.py
@@ -143,6 +143,77 @@ class Sigmoid:
     def _inverse(self, prop_correct: np.ndarray, threshold: np.ndarray, width: np.ndarray) -> np.ndarray:
         raise NotImplementedError("This should be overwritten by an implementation.")
 
+    def assert_sanity_checks(self, n_samples: int, threshold: float, width: float):
+        """ Assert multiple sanity checks on this sigmoid implementation.
+
+        These checks cannot completely assure the correct implementation of a sigmoid,
+        but try to catch common and obvious mistakes.
+        We recommend to use these for custom sigmoid subclasses.
+
+        The checks are performed on linear spaced stimulus levels between 0 and 1
+        and the provided sigmoid parameters.
+
+        Two checks for relations between parameters:
+          - `sigmoid(threshold_stimulus_level) == threshold_percent_correct`
+          - `|X_L - X_R| == width`
+            with `sigmoid(X_L) == alpha`
+            and  `sigmoid(X_R) == 1 - alpha`
+
+        Two checks for the inverse:
+          - `inverse(PC) == threshold_stimulus_level`
+          - `inverse(inverse(stimulus_levels) == stimulus_levels`
+
+        Two checks for the slope:
+          - `maximum(|slope(stimulus_levels)|)` close to `|slope(0.5)|`
+          - `slope(stimulus_levels) > 0`  (or < 0 for negative sigmoid)
+
+        Args:
+             n_samples: Number of stimulus levels between 0 (exclusive) and 1 for tests
+             threshold: Parameter value for threshold at PC
+             width: Width of the sigmoid
+        Raises:
+              AssertionError if a sanity check fails.
+        """
+        stimulus_levels = np.linspace(1e-8, 1, n_samples)
+        threshold_stimulus_level = threshold
+
+        # sigmoid(threshold_stimulus_level) == threshold_percent_correct
+        np.testing.assert_allclose(self(threshold_stimulus_level, threshold, width), self.PC)
+        # |X_L - X_R| == WIDTH, with
+        # with sigmoid(X_L) == ALPHA
+        # and  sigmoid(X_R) == 1 - ALPHA
+        prop_correct = self(stimulus_levels, threshold, width)
+        # When the sigmoid is negative, it is decreasing, so we compute the width on 1-prop_correct
+        # (Alternatively, we could have used `argmax` and swapped the indices)
+        if self.negative:
+            stimulus_levels = 1 - stimulus_levels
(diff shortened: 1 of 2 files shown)


psignifit/sigmoids.py
```diff
@@ -144,73 +139,6 @@ def _slope(self, stimulus_level: np.ndarray, threshold: np.ndarray, width: np.nd
     def _inverse(self, prop_correct: np.ndarray, threshold: np.ndarray, width: np.ndarray) -> np.ndarray:
         raise NotImplementedError("This should be overwritten by an implementation.")
 
-    def assert_sanity_checks(self, n_samples: int, threshold: float, width: float):
```
otizonaizit, line 147: I agree with moving this to the tests, but I think the purpose of this method was also to help the implementer of a custom sigmoid, i.e. a user. So you would write your own sigmoig by inheriting from this class and then run the superclass sanity checks to make sure that you indeed wrote something which looks like a sigmoid.

By moving the method to the tests we are hiding this from the user.

pberkes:
I'll implement the changes discussed in #141 

@otizonaizit regarding `assert_sanity_checks` ... would it be ok to move it to move it to a _function_ inside of `sigmoid.py`? instead of a method of `Sigmoid`

pberkes:
one clear issue that I want to solve, is that no matter how I set `self.PC`, all the tests always pass. This is because we test a round-trip, and as long as PC is the same for `_value` and `_inverse` it's going to pass

pberkes:
Any generic test that works for all sigmoids will not be able to verify that any specific sigmoid is implemented correctly. As a thought experiment, we could set all sigmoids to have the same implementation as `Gaussian`, and all tests would still pass

otizonaizit:
> @otizonaizit regarding `assert_sanity_checks` ... would it be ok to move it to move it to a _function_ inside of `sigmoid.py`? instead of a method of `Sigmoid`

yes, sure. As long as it discoverable by the user and we can document it, I don't think it's too relevant where it lives. We could even think of sticking it into `tools.py`, so that we know that is a collection of tools to be used by the user, and they will never need to know that there is a `sigmoid.py` module. OTOH, if they implement their own sigmoid, they'll have to read that code to see how it is done...

otizonaizit:
> Any generic test that works for all sigmoids will not be able to verify that any specific sigmoid is implemented correctly. As a thought experiment, we could set all sigmoids to have the same implementation as `Gaussian`, and all tests would still pass

But given that the sigmoids are defined by a mathematical formula, and that formula is implemented in `sigmoid.py`, I don't see how you want to test that they are implemented "correctly". Repeating the formula in the tests doesn't seem useful. So I think that this is something we can't test.
What the tests are for is verifying that changes in the API or typos don't break everything. Or at least, this is the way I understand these tests.

pberkes:
> But given that the sigmoids are defined by a mathematical formula, and that formula is implemented in `sigmoid.py`, I don't see how you want to test that they are implemented "correctly". Repeating the formula in the tests doesn't seem useful. So I think that this is something we can't test. What the tests are for is verifying that changes in the API or typos don't break everything. Or at least, this is the way I understand these tests.

@otizonaizit 
Computing the values by hand! see the new `test_sigmoid_values`

These tests are super important, in fact I JUST FOUND A BUG IN GAUSSIAN._SLOPE! I found it with the new slope test `test_sigmoid_slope`, and comparing the code with Matlab confirms it.

This was the old code (the bug is in the line for `normalized_stimulus_level`): 
```
    def _slope(self, stimulus_level: np.ndarray, threshold: np.ndarray, width: np.ndarray) -> np.ndarray:
        C = norminv(1 - self.alpha) - norminv(self.alpha)
        normalized_stimulus_level = (stimulus_level - threshold) / threshold * C
        normalized_slope = sp.stats.norm.pdf(normalized_stimulus_level)
        return normalized_slope * C / width
```

This is the Matlab code (https://github.com/wichmann-lab/psignifit/blob/master/getSlope.m#L35):
```
    case {'norm','gauss','neg_norm','neg_gauss'}   % cumulative normal distribution
        C         = my_norminv(1-alpha,0,1) - my_norminv(alpha,0,1);
        normalizedStimLevel = (stimLevel-theta0(1))/theta0(2).*C;
        slopeNormalized = my_normpdf(normalizedStimLevel);
        slope = slopeNormalized *C./theta0(2);
```
where  `theta0(1)` is the threshold and `theta0(2)` is the width. You can see that you need to divide by the width (which of course makes sense)

It's now fixed in this PR

otizonaizit:
> > But given that the sigmoids are defined by a mathematical formula, and that formula is implemented in `sigmoid.py`, I don't see how you want to test that they are implemented "correctly". Repeating the formula in the tests doesn't seem useful. So I think that this is something we can't test. What the tests are for is verifying that changes in the API or typos don't break everything. Or at least, this is the way I understand these tests.
> 
> @otizonaizit Computing some of the values independently, e.g. by hand! see the new `test_sigmoid_values`

**by hand**! Well, you deserve a medal for this: 🏅!

> These tests are super important, in fact I JUST FOUND A BUG IN GAUSSIAN._SLOPE! I found it with the new slope test `test_sigmoid_slope`, and comparing the code with Matlab confirms it.

good catch!!!
 
> This was the old code (the bug is in the line for `normalized_stimulus_level`):
> 
> ```
>     def _slope(self, stimulus_level: np.ndarray, threshold: np.ndarray, width: np.ndarray) -> np.ndarray:
>         C = norminv(1 - self.alpha) - norminv(self.alpha)
>         normalized_stimulus_level = (stimulus_level - threshold) / threshold * C
>         normalized_slope = sp.stats.norm.pdf(normalized_stimulus_level)
>         return normalized_slope * C / width
> ```
> 
> This is the Matlab code (https://github.com/wichmann-lab/psignifit/blob/master/getSlope.m#L35):
> 
> ```
>     case {'norm','gauss','neg_norm','neg_gauss'}   % cumulative normal distribution
>         C         = my_norminv(1-alpha,0,1) - my_norminv(alpha,0,1);
>         normalizedStimLevel = (stimLevel-theta0(1))/theta0(2).*C;
>         slopeNormalized = my_normpdf(normalizedStimLevel);
>         slope = slopeNormalized *C./theta0(2);
> ```
> 
> where `theta0(1)` is the threshold and `theta0(2)` is the width. You can see that you need to divide by the width (which of course makes sense).

and here yet again we learn how bad it is to rely on indexes instead of using a mapping `name:value` ;-)

> There were actually 2 bugs, the threshold in the Python version needs to be corrected in case PC is not 0.5
> 
> It's now fixed in this PR

The PR is still WIP? Ping here when it's time for a final review.

pberkes:
It's still WIP, I need to put the testing function back

psignifit/sigmoids.py
pberkes: moved to own function in same module

pberkes:
@otizonaizit @guillermoaguilar  after implementing your comments, this PR is finally ready to review

guillermoaguilar (approved):
It looks Ok to me.  Also I checked these changes with the example datasets, both the basic one and the negative slope one. Numbers match between matlab and python implementation (up to some differences in the marginals but that's  a different story).
I think it's good to go!

PR #149, opened 2024-09-25
Demo notebook for using psignifit in a log-space

pberkes:
I created a new demo notebook based on @lschwetlick 's "parameter recovery" one. 

This one shows how to run a fit in log-space, and how to transform threshold and width back to the original stimulus space.


otizonaizit:
The demo is quite cool! Thanks!

Don't you think it would make sense to use stimulus levels that are also _visually_ on a log scale? I mean instead of:
```python
stim_range = [2.0, 3.0]
threshold = 2.52
log_width = 0.25
stimulus_level = np.logspace(np.log10(stim_range[0]), np.log10(stim_range[1]), nsteps, base=10)
```
use something that spans two orders of magnitude or so? For example:
 ```python
stim_range = [1.0, 1000.0]
threshold = 100
log_width = 2
stimulus_level = np.logspace(np.log10(stim_range[0]), np.log10(stim_range[1]), nsteps, base=10)
```
and then plot using a semilog plot, for example:
```python
ax.semilogx(stimulus_level, perccorr, marker='o')
```
I think it makes the thing a bit more realistic?

pberkes:
good idea, I'll give it a go

pberkes:
@otizonaizit updated

otizonaizit:
LGTM!

Issue #92, opened 2024-09-19
support estimate types different from `MAP`? 


otizonaizit:
The Matlab version, and partially the master branch are supporting configuration['estimate_type'] different from `MAP`: `mean` and `MLE`. Do we want to support them? Felix says there's no need.

otizonaizit:
by the way `MAP` and `MLE` are exactly the same, and `mean` does not make sense: psignifit hasn't been even tested to work well with that estimate type

pberkes:
If implemented correctly, MAP should use the prior, while MLE should not

HeikoSchuett:
MLE and MAP do indeed result in the same behaviour in MATLAB. We got to many divergent results with the actual MLE to be useful.

The mean is sensible though and we did test it. It performs a little worse than the MAP, but as far as I know Bayesian stats usually considers the posterior mean to be a better summary.
Also computing the posterior mean once you have the marginals is ~5 lines of code, so it does not really seem like the place to cut...

pberkes:
As a far as I can tell, `MLE`, `MAP`, and `mean` are nowhere to be found in the code. The parameter `Configuration.estimate_type` is never used

otizonaizit:
Yes, that's the whole point of the discussion. The first porting from 
Matlab was done only for MAP. The question was if we want to _add_ 
support for other methods

8 Oct 2024 13:58:39 Pietro Berkes ***@***.***>:

PR #152, opened 2024-10-08
Remove estimate type parameter

pberkes:
First step for #92 : remove `estimate_type` parameter.

It was unused, and it was decided that we will rather *add* other types of parameter estimates to the Results object, rather than offering the estimates as alternative running modes.

The first commit removes the parameter, the second fixes typos I found while reading the demos


otizonaizit:
LGTM!

PR #153, opened 2024-10-08
Return the bounds dictionary in debug mode

guillermoaguilar:
This will allow to do the priors plot and the marginals plot with the correct range. These two plots only work in debug mode.

The plotting routines have been adjusted accordingly to use this new returned dict.

### DIFF
--- a/psignifit/psigniplot.py
+++ b/psignifit/psigniplot.py
@@ -1,12 +1,11 @@
 # -*- coding: utf-8 -*-
 from typing import Union, List
-
 import warnings
+
 import matplotlib.pyplot as plt
 import matplotlib.axes
 import numpy as np
 import scipy.stats
-from scipy.special import ndtri
 
 from . import psignifit
 from ._typing import ExperimentType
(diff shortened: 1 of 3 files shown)


psignifit/psigniplot.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from typing import Union, List
 
+import warnings
```
pberkes, line 4: nit: the import should be with the standard library import block (one line up)

pberkes (approved):
LGTM, tiny comment that can be ignored

psignifit/psigniplot.py
guillermoaguilar: out of curiosity, is there a guideline for that? order of imports? I never know.

pberkes: yes, from PEP8:
"""

Imports should be grouped in the following order:

- Standard library imports.
- Related third party imports.
- Local application/library specific imports.

You should put a blank line between each group of imports.
"""

guillermoaguilar: thank you darling.

PR #154, opened 2024-10-14
Fix `plot_prior` to plot 25 and 75 percentile of prior

### LINKED ISSUE #151: Bugs in prior plotting - selected sigmoids
The prior plotting function should add dots for the min, max, quantiles, and point estimate. However, they appear not at the right positions in the plot. 

@guillermoaguilar Could you please add the screenshots?

pberkes:
Fix #151 

The previous implementation computed the quantiles of the x-axis

There's a bit of noise because my editor also deleted unnecessary space characters -- sorry about that


otizonaizit:
LGTM!

PR #155, opened 2024-10-15
docs: adds warning on all pages again with proper syntax, also on generated examples

guillermoaguilar:


lschwetlick:
Looks good to me :)

PR #163, opened 2024-11-16
Save confidence intervals as a dictionary

### LINKED ISSUE #136: change format of confidence intervals structure
Currently confidence intervals are returned this way:
```python
>>> result.confidence_intervals['threshold'] 
[[0.0041551724137931035, 0.005110079575596817],
 [0.004218832891246685, 0.005110079575596817],
 [0.004346153846153846, 0.004982758620689655]]
```

I think it should be something like:

```python
>>> result.confidence_intervals['threshold'] 
{
 '95%' : [0.0041551724137931035, 0.005110079575596817],
 '90%' : [0.004218832891246685, 0.005110079575596817],
 '68%' : [0.004346153846153846, 0.004982758620689655]
}
```
which is way more explicit

### LINKED ISSUE #142: Remove the CI_method `stripes`
It's not implemented, anyway:

```
    elif mode == 'stripes':
        raise ValueError('Confidence mode "stripes" is not supported anymore. "'
                         f'"Use one of {list(CI_METHODS.keys())}.')
```
(from `confidence.py`, line 42)

pberkes:
Fix #136 
Fix #142 

Confidence intervals are not saved as 
```
>>> result.confidence_intervals['threshold'] 
{
 '0.95' : [0.0041551724137931035, 0.005110079575596817],
 '0.9' : [0.004218832891246685, 0.005110079575596817],
 '0.68' : [0.004346153846153846, 0.004982758620689655]
}
```

I propose to use '0.95' instead of '95%', because in my experience the latter would mean that we'll be writing forever code that converts '95%' into 0.95 and viceversa

### DIFF
--- a/docs/examples/demo_002.md
+++ b/docs/examples/demo_002.md
@@ -114,13 +114,6 @@ config['confP'] = [.95, .9, .68, .5]
 `CI_method` sets how the confidence intervals are computed in getConfRegion
 possible variants are:
 
-'stripes': Find a threshold with (1-alpha) above it
-     This will disregard intervals of low posterior probability and then move
-     in from the sides to adjust the exact CI size.
-     This can handle bounds and asymmetric distributions slightly better, but
-     will introduce slight jumps of the confidence interval when confp is
-     adjusted depending on when the gridpoints get too small posterior
-     probability.
 'project':
      Project the confidence region on each axis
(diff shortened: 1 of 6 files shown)


guillermoaguilar (approved):
This all look good to me.

Issue #143, opened 2024-09-24
Sometimes the confidence interval's upper bound is lower than the estimate


lschwetlick:
See the title.

See here for an example:
https://gist.github.com/lschwetlick/b6168ed60668076e0d0c510811c40fd3

This is only the case if you use the `percentiles` CI method.

otizonaizit:
@FelixWichmann : do you think this statistically possible and due to the imprecision of estimating the confidence intervals using the percentile method or is it the sign of a bug?

@guillermoaguilar : could you test if matlab shows he same effect using the same data?

FelixWichmann:
> @FelixWichmann : do you think this statistically possible and due to the imprecision of estimating the confidence intervals using the percentile method or is it the sign of a bug?
To me rather a sign of a bug, I'm afraid.

> @guillermoaguilar : could you test if matlab shows he same effect using the same data?
Indeed -- what does MATLAB show?

Sorry but I won't be able to look into psignifit issues until Friday late afternoon/weekend.
Please let me know if the problem persists.

Cheers, Felix

guillermoaguilar:
I ran the same data in MATLAB, it does give different results, the estimate is inside the CI. The MATLAB version estimates a much shallower function, and if you look at the marginals they are indeed different between the two implementations.

Here the plots from MATLAB and python (in that order) for the psychometric function:

![bug_ci_matlab1](https://github.com/user-attachments/assets/a31df248-4818-47b5-93d7-f6b074782ba7)
![bugci_python1](https://github.com/user-attachments/assets/fa9f8934-4c2f-4f26-8549-6c7aeb32b97e)

the marginal for the threshold:

![bug_ci_matlab2](https://github.com/user-attachments/assets/c0d3322a-0ebc-4ea1-8422-84b59875ff7d)
![bugci_python2](https://github.com/user-attachments/assets/343ad618-ff84-4e5f-92d0-53d226c7d75c)

the marginal for the width:

![bugci_python3](https://github.com/user-attachments/assets/785a25d6-7c02-47d1-88dc-4b12b919165f)
![bug_ci_matlab3](https://github.com/user-attachments/assets/c6c237d7-72ae-4161-a7a7-5f3ea9d74e45)


Now I have to say there are many things that make this a very special case
- very little data, just four x-values. Data matrix is: `data =  [[14. 14. 30.],  [24.  8. 30.],  [34. 28. 30.],  [44. 29. 30.]]`
[...]

PR #171, opened 2024-11-27
Remove obsolete configuration options

### LINKED ISSUE #145: Unused options in _configuration.py
While writing the docs for the options dict I found that some options in the configs are never used:

- fast_optim
- instant_plot
- grid_set_type

can they be removed?

otizonaizit:
Get rid of configuration options that are not used in the code.

This PR also removes the configuration option `stimulus_range`. For a rationale, see #143
The stimulus range is now only inferred from the data. Note that `width_min`, i.e. the lower bound of the width prior, can still be manually set. I am not sure if we should get rid of this too...

Fixes #145


pberkes:
Thank you @otizonaizit , the changes look good. If we get rid of `stimulus_range`, I think some of the notebooks in `demo` need to be modified.

Are we sure we want to remove it? In #143, the problem was that the minimum width was set in a naive way, and we have improved that. Felix' comment was that the experiment should normally cover the range of plausible stimuli, so setting it by hand is not useful. OTOH, there might be cases where it's not possible to cover the full range, and still want to infer a sigmoid on the full range?

otizonaizit:
btw, talking with a heavy user of psychometric function fitting libraries, I heard that it is difficult to imagine what sense an experiment would make where you do no sample the lower or higher part of the stimulus range but then want to fit a psychometric function and use that fit to claim something about thresholds and such...
If it is only about synthetic data, then we can keep the option internally as private but there's not need to advertise it as something useful. My fear is that people will follow the example codes where it is used and do the same mistake we were doing in #143 ...

pberkes:
alright, let's do it then

pberkes:
There's one last thing to do: remove `options["stimulus_range"] = stim_range` from the notebook `psignifit/demos/parameter_recovery_demo.ipynb`. It appears in two places. In principle, the rest should continue working as intended.

otizonaizit:
@guillermoaguilar : before proceeding with the removal of `stimulus_range` I'd like to get your OK too.

guillermoaguilar:
Yes sorry for the silence. 
`stim_range` is not really needed from the user perspective IMO. 
It should be private.


> There's one last thing to do: remove `options["stimulus_range"] = stim_range` from the notebook `psignifit/demos/parameter_recovery_demo.ipynb`. It appears in two places. In principle, the rest should continue working as intended.

In the rearrangement of documentation I've already removed it. But I haven't merge that yet. Leave that to me.

guillermoaguilar (approved):

otizonaizit:
> There's one last thing to do: remove `options["stimulus_range"] = stim_range` from the notebook `psignifit/demos/parameter_recovery_demo.ipynb`. It appears in two places. In principle, the rest should continue working as intended.

well its' not so easy: that demo was already broken by #161 :

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[7], line 1
----> 1 assert np.isclose(res.parameter_estimate['lambda'], lambda_, atol=1e-4)

AttributeError: 'Result' object has no attribute 'parameter_estimate'
```

I think that the demos need to be checked again. I don't think it is a good idea for me to fix it in this PR, given that the thing is already broken...
By the way, it may make sense to still have a `Result.parameter_estimate` which is an alias of `Result.parameter_estimate_MAP`?

guillermoaguilar:
> I think that the demos need to be checked again. I don't think it is a good idea for me to fix it in this PR, given that the thing is already broken... By the way, it may make sense to still have a `Result.parameter_estimate` which is an alias of `Result.parameter_estimate_MAP`?

Yes, it should be an alias for _MAP.
It's too complicated to ask the user to decide if to use `parameter_estimate_MAP`  or something else.... `parameter_estimate` should be the default.

guillermoaguilar:
Then I guess I should merge the documentation changes and see what's broken.

guillermoaguilar (approved):

otizonaizit:
OK, I made `stimulus_range` private. Which means that it can still be used internally with the name `Configuration._stimulus_range` but it is not advertised anywhere.

PR #177, opened 2024-12-04
Make sure that the sigmoids can handle the case where alpha==0.0

### LINKED ISSUE #175: standard_parameters will fail when alpha==0
I forgot to test for the corner case / not unlikely case when alpha is exactly 0.0 .
The standard_parameters method will fail for those cases, because the rescaling will give inifinities and nans.

other stuff would have failed also before #169, for example in the Logistic sigmoid we had
`C = 2 * np.log(1 / self.alpha - 1) / width`
which is going to cause a division by zero.

pberkes:
Fix #175

### DIFF
--- a/psignifit/sigmoids.py
+++ b/psignifit/sigmoids.py
@@ -242,7 +242,10 @@ class Gaussian(Sigmoid):
         return scipy.stats.norm, {}
 
     def _standard_parameters(self, threshold: N, width: N) -> list:
-        scale = width / (self._ppf(1 - self.alpha) - self._ppf(self.alpha))
+        if self.alpha == 0.0:
+            scale = width
+        else:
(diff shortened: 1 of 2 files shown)


guillermoaguilar:
It's fine for me, although alpha=0 means that the width is from -inf to inf, something that really doesnt make sense in practice. 

You either take the default alpha=0.05, so width is between Pc 0.05 and 0.95
or you do something even more narrow, say alpha = 0.1 so Pc 0.1 - 0.9
or something standard, like standard deviation in the case of Gauss.

So corner case of alpha=0 shouldnt really come. Or maybe I'm missing something?

otizonaizit:
So @guillermoaguilar you think it make more sense to let the code fail for `alpha==0`? In the sense that this can only happen if the user has made a mistake and they deserve an `Exception`?

otizonaizit:
OK, closing as requested

Issue #180, opened 2024-12-05
demos location 


guillermoaguilar:
The folder psignifit/demos has some demos that are now redundant with the jupyter book documentation in docs/ 
These are: 
- demo_001 to _006
- parameter_recovery_demo.ipynb  and
- plot_all_sigmoids.py

All these have been migrated to the markdown format and reorganized in docs/. 

Now the question is to whether delete these ones there in psignifit/demos or not. 

These get executed when the documentation is built. But the build does not raise errors during CI testing. The only way of checking is manually looking.

So it is confusing to have them in two places, but at the same time the documentation in docs/ can be broken and we dont notice, because tests still pass.
 
So I dont know, what do you all think?

otizonaizit:
I just checked and it is possible to configure `jupyter-book` to exit with an error if one of the generated notebooks fails to execute:

```python
jupyter-book build --warningiserror docs/
```

so, we just need to add here: https://github.com/wichmann-lab/python-psignifit/blob/710f9f7db6992ece9f5aae9e5a096c6b562fb36f/.github/workflows/build-docs.yml#L38

guillermoaguilar:
Ok that's good! I'll take care of it then

PR #181, opened 2024-12-09
Exit with an error when building docs if a notebook fails to execute

otizonaizit:
After merging this PR the CI will fail if a notebook in the docs fails to execute. This way we can be sure that the docs are always up-to-date with the latest changes in the code.

To achieve this I had to adjust a bit the doc-building procedure.

Important: 
- I disabled the building of https://github.com/wichmann-lab/python-psignifit/blob/main/docs/examples/priors.md because it is still using the `stimulus_range` option that was removed in #171 .
- I removed the obsolete demo notebooks as discussed in #180 

Fixes: #180


guillermoaguilar:
All good except that this PR is deleting the demos

- parameter_recovery_in_logspace_demo.ipynb
- parameters_estimates.ipynb

But these have not been migrated yet.

What you can do is just move them to docs/to-be-migrated   or something
and then add this folder to be excluded from docs build

psignifit/demos/parameter_recovery_in_logspace_demo.ipynb
guillermoaguilar, line 1: keep this one, move to docs/not-yet-migrated

psignifit/demos/parameters_estimates.ipynb
guillermoaguilar, line 1: this one too

otizonaizit:
> All good except that this PR is deleting the demos
> 
>     * parameter_recovery_in_logspace_demo.ipynb
> 
>     * parameters_estimates.ipynb
> 
> 
> But these have not been migrated yet.

OK, I reinstated the two non ported demos

guillermoaguilar:
LGTM!

PR #184, opened 2024-12-16
Move rescaling of psychometric function related to lambda and gamma in `Sigmoid` object

### LINKED ISSUE #170: Sigmoid.__call__ does not take `lambda` and `gamma`
At the moment, `Sigmoid.__call__` does not take the parameters `lambda` and `gamma`. It is odd, because the other public methods, `Sigmoid.slope` and `Sigmoid.inverse` to take those parameters into account.

The smell comes out in `psignifit.tools.psychometric`, which does take those parameters, and where the values are computed by calling `Sigmoid`, then rescaling:
```
    sigmoid = sigmoid_by_name(sigmoid_name)
    sigmoid_values = sigmoid(stimulus_level, threshold=threshold, width=width)
    psi = gamma + (1.0 - lambda_ - gamma) * sigmoid_values
```

I think this simple operation should be performed in `Sigmoid.__call__`

pberkes:
Fix  #170 

See #170 for details, but in short there were two smells in the code: 1) every call to Sigmoid.__call__ was followed by a rescaling using lambda and gamma; 2) all other public methods of Sigmoid take lambda and gamma as parameters, except for __call__.

I moved the rescaling inside Sigmoid.__call__ and got rid of both

### DIFF
--- a/psignifit/_posterior.py
+++ b/psignifit/_posterior.py
@@ -146,7 +146,6 @@ def log_posterior(data: np.ndarray, sigmoid: Sigmoid, priors: Dict[str, Prior], 
 
     if gamma is None:
         gamma = lambd
-    scale = 1 - gamma - lambd
 
     pbin = 0
     p = 0
@@ -159,7 +158,7 @@ def log_posterior(data: np.ndarray, sigmoid: Sigmoid, priors: Dict[str, Prior], 
         if trials == 0:
             continue
 
-        psi = sigmoid(level, thres, width) * scale + gamma
+        psi = sigmoid(level, thres, width, gamma=gamma, lambd=lambd)
 
         # Separate cases to avoid warnings problems with np.log(...)
         # for correct_trials == 0 or (trials - correct_trials) == 0
--- a/psignifit/psigniplot.py
+++ b/psignifit/psigniplot.py
@@ -56,8 +56,13 @@ def plot_psychometric_function(result: Result,  # noqa: C901, this function is t
     x = np.linspace(x_data.min(), x_data.max(), num=1000)
     x_low = np.linspace(x[0] - extrapolate_stimulus * (x[-1] - x[0]), x[0], num=100)
     x_high = np.linspace(x[-1], x[-1] + extrapolate_stimulus * (x[-1] - x[0]), num=100)
-    y = sigmoid(np.r_[x_low, x, x_high], params['threshold'], params['width'])
-    y = (1 - params['gamma'] - params['lambda']) * y + params['gamma']
+    y = sigmoid(
+        np.r_[x_low, x, x_high],
+        threshold=params['threshold'],
+        width=params['width'],
+        gamma=params['gamma'],
+        lambd=params['lambda'],
+    )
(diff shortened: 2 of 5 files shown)


otizonaizit:
Cool! Thanks! What about getting rid of [`tools.psychometric`](https://github.com/wichmann-lab/python-psignifit/blob/b97791068104a55db0dfccd401fe243ec52f6a16/psignifit/tools.py#L54) That function ow is basically `Sigmoid.__call__`. I think it could be done in this PR directly.

Regarding [`tools.psychometric_with_eta`](https://github.com/wichmann-lab/python-psignifit/blob/b97791068104a55db0dfccd401fe243ec52f6a16/psignifit/tools.py#L88) : maybe now that `Sigmoid` does the right thing this function is not needed anymore or it needs to become `sigmoid_with_eta`? what do you think @pberkes ?

pberkes:
My opinion:
- `psychometric_with_eta` is still needed, Sigmoid does not know about over-dispersion (and shouldn't)
- the name `psychometric_with_eta` I think it's correct: the intention is to get the values of a psychometric function, which incidentally is a sigmoid
- I have mixed feelings about those functions... they are useful utilities for the demos and especially the tests. I think they should also (or optionally) sample data from the psychometric function. At a minimum, I would keep `psychometric` as a test utility, otherwise we'll duplicate those two lines 13 times in the tests. Maybe they should be moved somewhere more discrete? e.g. `testing_tools`?

guillermoaguilar (approved):
LGTM

guillermoaguilar:
On second thought, I think

- `psychometric_with_eta` should stay, as Pietro says, eta is not a parameter for `Sigmoid` and shouldnt. Eta is part of the beta-binomial model underlying estimation.

- `psychometric` seems redundant in terms of the code, but I think it's useful as a nice function wrapper for sigmoids. In some demos it is used to demonstrate stuff... for example `parameter_recovery_demo.md` in docs/examples. I find it a nice tool to have, it's easy to use in the syntax. Maybe we can improve the docstrings of `psychometric` and write that is just a convenience function, particularly for simulations.

pberkes:
@guillermoaguilar  I improved the docstrings
@otizonaizit  any thought?

lschwetlick:
LGTM. I agree about keeping ´psychometric_with_eta´ around. I've found it useful for the demos.

otizonaizit:
well, OK. I still think that the two lines of `psychometric` – which by the way in tests could be reduced to one, because there's no need to call the sigmoid by their string-name (we could use the class objects directly instead) – those lines are still not worth a function.

When used to generate data, I think the audience of such demos is an audience which will not be confused by naming a thing sigmoid instead of psychometric, but that is my personal view.

I agree with the case being made for `psychometric_with_eta`, but that does not require `psychometric` to survive ;-)

But I don't want to stay in the way, so feel free to merge: my dissent will be recorded for ever in this issue :-)))

pberkes:
@otizonaizit  let's do it right: I'll merge the PR, as it fulfills the issue it set out to solve. I openend #189 for the removal of `psychometric`, for discussions. I'll draft a PR and we can see what it looks like

PR #186, opened 2024-12-16
Fixes wrong position of black dot in prior plot

### LINKED ISSUE #151: Bugs in prior plotting - selected sigmoids

guillermoaguilar:
a residual difference between the MATLAB plot and the python plot.
Fixes #151

### DIFF
--- a/psignifit/psigniplot.py
+++ b/psignifit/psigniplot.py
@@ -329,8 +329,9 @@ def plot_prior(result: Result,
         prior_cdf = np.cumsum(prior_val * prior_w)
         q25_index = np.argmax(prior_cdf > 0.25)
         q75_index = np.argmax(prior_cdf > 0.75)
+        prior_mean = np.sum(prior_x * prior_val)/np.sum(prior_val)  # ? copied from MATLAB
 
-        x_percentiles = [estimate[param],
+        x_percentiles = [prior_mean,
                          min(prior_x),
                          prior_x[q25_index],
                          prior_x[q75_index],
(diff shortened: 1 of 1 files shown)


pberkes:
My bad, I had misread the Matlab code. The black dot really is the mean of the prior... a bit strange, plotting the 0, 25, 75, and 100 percentiles but then the mean and not the median. I also thought that plotting the result of the psychometric fit was a smart thing to do, to check that it is not in a region with very low prior probability.

Anyway, this is what Matlab does, so 👍

psignifit/psigniplot.py
```diff
@@ -329,8 +329,9 @@ def plot_prior(result: Result,
         prior_cdf = np.cumsum(prior_val * prior_w)
         q25_index = np.argmax(prior_cdf > 0.25)
         q75_index = np.argmax(prior_cdf > 0.75)
+        prior_mean = np.sum(prior_x * prior_val)/np.sum(prior_val)  # ? copied from MATLAB
```
pberkes, line 332: Please remove the comment. I think the division by `np.sum(prior_val)` is unnecessary, because the prior functions should be normalized already. Let's leave it in, for consistency with Matlab

guillermoaguilar: comment removed!

I tried without the division but it did not work, to be honest I don't really understand the operation, I just copied the line from matlab. It seems to work.

The black dot should be the 'prior mean', it is also described  in the paper like that.

PR #187, opened 2024-12-16
Adds 2D marginals also for fixed params

### LINKED ISSUE #158: Correct figure plot_bayes
Plot_bayes:

plot_bayes

- [x] bug: the y-axis in each 2D colormap is swapped.

and make it more similar to MATLAB version:

- [x] the 2D colormap for a fixed parameter does not exist, instead a placeholder text was added saying this parameter was fixed. To replicate it we need to repeat the marginal for the non-fixed parameter

- [ ] use same colormap

MATLAB version:
demo_005_4

guillermoaguilar:
This copies the behavior of the MATLAB version, in which the marginal is simply repeated along the singleton axis so a heatmap can still be drawn.

This fixes #158

### DIFF
--- a/psignifit/psigniplot.py
+++ b/psignifit/psigniplot.py
@@ -376,13 +376,41 @@ def plot_2D_margin(result: Result,
     other_param_ix = tuple(i for param, i in parameter_indices.items()
                            if param != first_param and param != second_param)
     marginal_2d = np.sum(result.debug['posteriors'], axis=other_param_ix)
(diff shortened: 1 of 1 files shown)


guillermoaguilar:
new plot looks [like this](https://psignifit--187.org.readthedocs.build/en/187/examples/plotting.html#plot-all-2d-posteriors)

90018dafb0be130d3991eb238bf038da5b2f398e616c40b901e124fb89db807c

pberkes:
LGTM!

Issue #189, opened 2024-12-18
Remove `psychometric` utility function


pberkes:
It is a two-liner used in tests and demos, see discussion in #184

otizonaizit:
Fixed by #190

PR #190, opened 2024-12-18
Remove `tools.psychometric` function

pberkes:
This is what #189 would look like... thoughts?


otizonaizit:
It looks good to me and much less dramatic than I expected after the discussion in #184 :)
Thanks!

tests/test_param_recovery.py
```diff
@@ -4,13 +4,14 @@
...
 from psignifit import psignifit
-from psignifit.sigmoids import ALL_SIGMOID_NAMES
-from psignifit.tools import psychometric, psychometric_with_eta
+from psignifit.sigmoids import ALL_SIGMOID_NAMES, Gaussian, sigmoid_by_name
+from psignifit.tools import psychometric_with_eta
 from psignifit._utils import fp_error_handler
 
-@pytest.mark.parametrize("sigmoid", list(ALL_SIGMOID_NAMES))
+
+@pytest.mark.parametrize("sigmoid_name", list(ALL_SIGMOID_NAMES))
```
otizonaizit, line 12: I just realized that by using `ALL_SIGMOID_NAMES` we are testing the same sigmoid objects over and over again, because some names are just aliases. To test each sigmoid class only once we would need to do something like:
```python
from psignifit.sigmoids import _CLASS_BY_NAME

ALL_SIGMOID_CLASSES = set(_CLASS_BY_NAME.values())
```
@pberkes : do you think this fix fits in this PR or do we need a new one?

otizonaizit (changes requested):

tests/test_param_recovery.py
pberkes: I realized that was the case, as well, but since the names are mainly what is exposed to the user, I was fine with it.

If anything, I would put `ALL_SIGMOID_CLASSES` in `sigmoids.py`. I'm +0 on this one

otizonaizit: well, but now we are wasting a lot of time testing the same thing over and over again: the tests are quite slow already, no? I don't see why it is useful to test the names instead of testing the classes directly. We can have a test to verify that each name corresponds to a class or so, if you want the user API to be tested.

pberkes: @otizonaizit  I'm not sure if @lschwetlick and @guillermoaguilar  will like this PR... if they like it, I'm happy to do it here, otherwise it needs to be done in a new PR

pberkes:
```
ALL_SIGMOID_NAMES = set(_CLASS_BY_NAME.keys())
ALL_SIGMOID_NAMES |= {'neg_' + name for name in ALL_SIGMOID_NAMES}
```

otizonaizit: well, that's easily solved: parametrize over the class names, and then run the test once with the class name initialized with `negative=False` and then with the same class name, but initialized with `negative=True`. Or even nicer, parametrize also over the two values of `negative`.

guillermoaguilar: really here I dont have a strong opinion ;) 
but I guess what @otizonaizit  suggest makes sense... test the classes directly *and* then test that the names map to the classes correctly.

PR #192, opened 2024-12-21
Updates documentation - Priors

guillermoaguilar:
updates documentation on priors

- explanation default priors
- how to change stimulus range for adaptive procedures
- pass a custom prior function
- check information against matlab wiki

updates also function that plots psychometric function,

- adds a parameter to scale the marker size for the data

### DIFF
--- a/docs/examples/priors.md
+++ b/docs/examples/priors.md
@@ -61,19 +61,27 @@ res = ps.psignifit(data, sigmoid='norm', experiment_type='2AFC', debug=True)
 psigniplot.plot_prior(res)
 ```
 
-You should check that the assumptions `psignifit` take
-are actually true in the case of your data.
-e.g. check whether one of the following statements holds:
+The panels in the top row show the prior densities for threshold, width
+and lapse rate respectively. In the second row psychometric functions
+corresponding to the 0, 25, 75 and 100% quantiles of the prior are
+plotted in colour. The other parameters of the functions are left at
+the prior mean value, which is also marked in all panels in black.
+As an orientation the small black dots in the lower panels mark the
+levels at which data were sampled.
+
+
+You should check that the assumptions we make for the heuristic to
+work are actually true in the case of your data.
+e.g. check whether (at least) one of the following statements holds:
+
 
 - You understand what our priors mean exactly and judge them to be appropriate.
 - You are sure you recorded a trial well above and a trial well below threshold.
-- Your posterior concentrates on an area for which the prior was constant.
+- Your posterior concentrates on an area for which the prior was (nearly) constant.
 
-You can evaluate your priors with above plot. In the upper panels the plots show the priors for the `threshold`, `width` and $\lambda$ parameters.
-Five selected values are shown color-coded, as points on the prior and below the corresponding psychometric functions.
-The black case correspond to the mean of the prior; the yellow, red, blue and green cases correspond to the minimum, 25 % percentile, 75 % percentile, and maximum of the prior, respectively.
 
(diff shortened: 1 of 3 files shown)


docs/examples/priors.md
```diff
@@ -64,18 +61,41 @@ res = ps.psignifit(data, sigmoid='norm', experiment_type='2AFC', debug=True)
...
+- Your posterior concentrates on an area for which the prior was (nearly) constant.
+
+
+You can evaluate your priors with above plot. With these color-coded sigmoids
+you can evaluate the adequacy of the prior:
+
+For a threshold prior to be adequate (first column), it should be flat along all your stimulus levels; this will ensure that the threshold estimation is driven exclusively by the data. The stimulus levels of your data are shown as black-dots in the x-axis.
+The prior is flat (constant) for all threshold values between the red and blue sigmoids; both are equally likely. The yellow and green sigmoids are very unlikely, and this makes sense as they are outside of the range of stimulus levels. Here you can see that `psignifit` assumes that you choose stimulus values along the whole range the psychometric function.
+
+Similarly, for the width prior (middle column) the sigmoids between varing between the red and blue cases are more or less likely; contrarily the yellow and green sigmoids are very unlikely. Again this make sense, as `psignifit` assumes that you sampled the whole range of the psychometric function and not just around the threshold.
```
pberkes, line 89: a "between" too many

pberkes: this "makes" sense

pberkes: I fixed these two typos myself

pberkes:
@otizonaizit  this page shows that in some real-life cases it makes sense to adjust the stimulus range

pberkes:
LGTM

otizonaizit:
> @otizonaizit this page shows that in some real-life cases it makes sense to adjust the stimulus range

Hmm, and yet I fail to make sense of that example. I am not enough into the matter to judge what it means that with an "adaptive" procedure where you don't test stimuli below threshold, by setting the range of the prior you magically get more info about the region you didn't test... I really think this is a discussion to be held again with the experts. @FelixWichmann : what do you think of this example? Is it sound?
If this thing makes sense, then `stimulus_range` must be public.

guillermoaguilar:
> > @otizonaizit this page shows that in some real-life cases it makes sense to adjust the stimulus range
> 
> Hmm, and yet I fail to make sense of that example. I am not enough into the matter to judge what it means that with an "adaptive" procedure where you don't test stimuli below threshold, by setting the range of the prior you magically get more info about the region you didn't test... I really think this is a discussion to be held again with the experts. @FelixWichmann : what do you think of this example? Is it sound? If this thing makes sense, then `stimulus_range` must be public.

This documentation page is just a mirror copy of the one for the MATLAB version 

https://github.com/wichmann-lab/psignifit/wiki/Priors

I just adapted the syntax and added some more text to explain the graphs. So I think it is good as it is.

And FYI, actually adaptive procedures are very commonly used, instead of testing stimulus levels in the whole range. People do this to save time, but at a cost. After they collect the data they still want to fit a psychometric function. The problem is that usually the data does not cover the whole function, it tends to be on the higher side, simply  because this is how the adaptive procedure works: roughly speaking, it presents you a high stimulus value, if you correctly answer after X times it lowers the value, and repeat this until you make mistakes... then the procedure makes task easier by increasing the stimulus value. After X number of 'reversals' it stops. 
Then with such a procedure is very likely that you don't sample the whole range of the psychometric function.

I'd say then we will need the `stimulus_range` to be public... this wasn't clear to me before when we removed it :/

PR #193, opened 2024-12-21
Parametrize tests using the sigmoid classes, not the names

pberkes:
... because the names are sometimes aliases for the same class, as discussed in PR #190


otizonaizit (approved):
LGTM, thanks @pberkes : good and compact!

Issue #83, opened 2024-09-19
Update documentation


pberkes:

lschwetlick:
in the file `index.rst` it says:
'If you require additional information, or to report errors or questions please contact us: ' and then gives the contact of Felix and Heiko. Should this be changed?

lschwetlick:
There is also an Acknowledgements section there. Should we acknowledge the members of this sprint?

lschwetlick:
I wonder whether we should keep this style of always referencing the Matlab Version. Or whether it might be better to copy the text and adapt it rather than point out the differences. For a new user who is not migrating from matlab I imagine it would be kind of annoying to be confronted with this comparison exercise...

Other ToDos:
- [ ]  fix `test_MATLAB_version.rst`
- [ ] `getThreshold` and `getSlope` and `getSlopePC` as described in file `docs/how_to/How-to-Get-Thresholds-and-Slopes.rst` dont exist
- [ ] the option to set the threshold does not exist as described in file `docs/how_to/How-to-Change-the-Threshold-Percent-Correct.rst`
- [ ] the user guides page is unlinked in the TOC on the left (https://psignifit.readthedocs.io/en/latest/user_guides/#user-guides)

otizonaizit:
> I wonder whether we should keep this style of always referencing the Matlab Version. Or whether it might be better to copy the text and adapt it rather than point out the differences. For a new user who is not migrating from matlab I imagine it would be kind of annoying to be confronted with this comparison exercise... This is particularly the case in all the `user_guides`.
> My personal vote is to get rid of the references to the matlab version, and copy paste over all the relevant stuff. Thoughts?

I agree that the references to the MATLAB version are confusing. We can refer to it right at the beginning once, linking to their docs. And if deemed necessary, we can write a page guiding the user in the transition from the MATLAB version to the Python version. But the default in my opinion should be to assume that the user is coming directly to our version.
 

> Other ToDos that need some decisions before fixing:
> 
>     * [ ]   fix `test_MATLAB_version.rst`. How far are we with decisions about matlab tests?
this may take still some time, I fear.

>     * [ ]  `getThreshold` and `getSlope` and `getSlopePC` as described in file `docs/how_to/How-to-Get-Thresholds-and-Slopes.rst` dont exist. Should we implement them or not document?

I thought that these are methods on the sigmoid objects, and should be already obtainable from the results object, see for example #137 ? Or am I misunderstanding something?

[...]

PR #199, opened 2024-12-25
Update Docs: Pooling

lschwetlick:
I fixed the page for the Pooling utility in the docs.

There was a strange bug (?) in the `plot_psychometric_function` code which reversed the logic of smaller dots representing blocks with fewer stimulus levels. Its strange because it seemed very intentional. I changed it here, but if you can see any reason not to please let me know. That function was recently touched by @guillermoaguilar and @pberkes but the line seems to have been there since always. 

Adresses issue #83


docs/examples/pooling_utilities.md
```diff
@@ -0,0 +1,96 @@
...
+
+Psignifit comes with an automatic pooling utility, which can pool trials of very similar stimulus levels into blocks to reduce the time required for fitting, to profit from the beta-binomial model and to improve interpretability of data-plots.
+
+In contrast to the matlab implementation, python-psignifit does not pool implicitly. Instead a warning is printed, if pooling might be useful. Then pooling can be run as a separate call using psignifit.tools.pool_blocks
+
+This way of pooling data is presumably acceptable for many datasets, but pooling by hand is probably more exact.
+
+```{code-cell} ipython3
+
+#act = pool_blocks(inp, max_tol=0.4, max_gap=10, max_length=10)
```
pberkes, line 25: This line is commented out, is it meant to show how the utility is called? Maybe introducing it with "the general form of the manual pooling utility looks like this:" might make sense

docs/examples/pooling_utilities.md
```diff
@@ -0,0 +1,96 @@
...
+pooled = pool_blocks(data, max_tol=0.01, max_length=25)
+print(pooled)
+res = ps.psignifit(pooled, **options)
+fig, ax = plt.subplots(1, 1);
+psigniplot.plot_psychometric_function(res, ax=ax);
+```
+
+This brakes the large blocks up again allowing us to notice if there was more variability over time than expected.
+
+The last option gives us a different rule to achieve something in a similar direction: We can enforce that a block must be collected en bloc like this:
```
pberkes, line 86: I'm not sure I got this: a block can be of unlimited size, and there is a maximum gap of 0 between blocks? Is it not like the first case without options?

psignifit/psigniplot.py
```diff
@@ -49,8 +48,8 @@ def plot_psychometric_function(result: Result,  # noqa: C901, this function is t
     if plot_data:
         y_data = data[:, 1] / data[:, 2]
         # the size is proportional to the sqrt of the data size, as in the MATLAB version.
-        # We added a factor of 1000 to make visually similar to the MATLAB version
-        size = np.sqrt(_data_size / data[:, 2])*(data_size*1000)
+        # We added a factor of 100 to make visually similar to the MATLAB version
+        size = np.sqrt(data[:, 2])*(data_size*100)
```
pberkes, line 52: As far as I know, the factor of 100-1000 was always there, and recently @guillermoaguilar added a `data_size` option in case this default is not good enough. Maybe we should move the factor directly in the keyword argument? I.e. make `data_size = 100`  by default, and then the user can adjust it if necessary

pberkes (changes requested):
Thank you, I added a could of minor comments.

Please also remove the file `user_guides/Pooling-Utility` , and its reference in `_toc.yml`. It refers to a work-in-progress page.

psignifit/psigniplot.py
pberkes: oh but I see now the "reversing" of the logic... it's true, it seems the wrong way around

lschwetlick: `data_size` and `_data_size` were two different values though. That was causing the problem.

pberkes: true, sorry I missed that

lschwetlick: its true that we could abandon the 100 scaling factor... on the other hand its also intuitive to have the input value 1 be a reasonable size already. I dont mind either way.

docs/examples/pooling_utilities.md
lschwetlick: The first one pools all trials within a tolerance. The next one pools only up to a block size of 25. The last one pools only subsequent trials. If I understand correctly.

docs/examples/pooling_utilities.md
lschwetlick: done

docs/examples/pooling_utilities.md
pberkes: got it, thanks. maybe "we can enforce to pool only subsequent trials", instead of "must be collected en block"?

psignifit/psigniplot.py
pberkes: no need for changes, afaic

pberkes:
LGTM

lschwetlick:
Ok, fixed the phrasing on the one remaining point. Good to merge from my side!

docs/examples/pooling_utilities.md
lschwetlick: ok

pberkes (approved):

pberkes:
The changes look good. The log-space notebook now has a plot that looks like below for some reason... would mind fixing it please? then it can be merge, as far as I'm concerned


image

PR #200, opened 2024-12-25
Update Docs: advanced options page

lschwetlick:
I started fixing the advanced options page for #83. I already made this PR now because I will have some questions about some of the options, and some clearly dont work yet, like alpha_width.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -20,7 +20,7 @@ latex:
 html:
   home_page_in_navbar         : true
   use_repository_button       : true  # Whether to add a link to your repository button
-  announcement                : "Some parts of this documentation are still work in progress! Some information might be outdated." # A banner announcement at the top of the site.
+  # announcement                : "Some parts of this documentation are still work in progress! Some information might be outdated." # A banner announcement at the top of the site.
 
 # Add a bibtex file so that we can create citations
 #bibtex_bibfiles:
--- a/docs/examples/advanced-options.md
+++ b/docs/examples/advanced-options.md
@@ -71,7 +71,6 @@ In this case the lapse rate is sampled at 50 points giving you a much more exact
 options['grid_steps'] = {'lambda': 50}
 res = ps.psignifit(data, **options)
 ```
-#TODO: this doesnt seem do anything
 
 ## Confidence intervals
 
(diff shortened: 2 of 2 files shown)


lschwetlick:
So I just checked the options and it turns out there's 2 related ones, `thresh_PC` and `width_alpha`, that are not implemented. `thresh_PC` at least throws a reasonable error, changing `width_alpha` gives something cryptic.

Should we implement them still, document their non-implementation, or remove the option entirely?
@pberkes @otizonaizit

otizonaizit:
> So I just checked the options and it turns out there's 2 related ones, `thresh_PC` and `width_alpha`, that are not implemented. `thresh_PC` at least throws a reasonable error, changing `width_alpha` gives something cryptic.
> 
> Should we implement them still, document their non-implementation and make the errors useful, or remove the option entirely?
> @pberkes @otizonaizit

I'll need some time to look into it. But in the meanwhile: would you find these parameters useful as an experimentalist? I think this is the first question to answer.

lschwetlick:
Personally, I haven't had a usecase for them yet, but I also only just recently turned psychophysicist. Maybe @FelixWichmann has an opinion?

lschwetlick:
This is the last TODO for the docs, so for the release we just need to decide what to do.

pberkes:
@lschwetlick  `thresh_PC` and `width_alpha` are used in `Configuration.make_sigmoid`, and from there in several other places, including the fitting function `psignifit` itself. As far as I understand, they will be rarely changed, but they are very integrated with the code right now. What kind of error do you get?

pberkes:
@lschwetlick  sorry I had missed @guillermoaguilar 's comment, https://github.com/wichmann-lab/python-psignifit/issues/83#issuecomment-2370969018

and the places where it fails is in `_priors.default_prior`:

```
    if not np.isclose(threshold_percent_correct, 0.5):
        raise ValueError("Default prior 'threshold' expects thresh_PC=0.5, got {thresh_PC = }")
```

damn

lschwetlick:
for thresh_PC:
```
ValueError: Default prior 'threshold' expects thresh_PC=0.5, got {thresh_PC = }
```
for width_alpha:
```
AssertionError: Prior 'width' returns zeros.
```

pberkes:
it could be pushed to v4.4?

pberkes:
regarding this PR: I suggest to remove the "under construction" banner in the documentation, and for `width_alpha` and `thresh_PC` write a description in the documentation, with a warning that at the moment they cannot be changes from the default. Then there is a failure when building the docs to be fixed, and this PR can be merged.

Fixing those parameters requires anyway a separate PR, we can decide if doing that for 4.4 or we have time before.

PR #201, opened 2024-12-25
Restore stimulus range as a public configuration variable

otizonaizit:
As discussed in #192 , the configuration option `stimulus_range` needs to be public and advertised.
I personally still think this is a statistical trick and I think the option can and will be abused,  but the option is there for people that know what they are doing (or think they know ;-)))

This PR partly reverts #171


pberkes:
LGTM

lschwetlick:
LGTM

PR #202, opened 2024-12-26
Add a docs page for the get_standard_parameters method.

lschwetlick:
I tried to describe the function as best I knew how. Should we have a more in depth demonstration and tutorial, or do you think this is fine?


pberkes:
LGTM! there's an error with building the docs to fix

pberkes:
two errors actually: 1) `res` does not exist in the first cell; 2) it is called `result` in the second cell

lschwetlick:
eh, yes. So the thing is that in this case the code cells just shouldnt be executed. Fixed it!

lschwetlick:
Ready to merge from my side then.

pberkes:
thanks

Issue #204, opened 2024-12-26
Options `thresh_PC` and `width_alpha` are still not working properly


otizonaizit:
See discussions in #200 and #83.

pberkes:
Regarding `width_alpha`:

-  as far as I can tell, `width_alpha` is correctly used to initialize the prior of `width`, and in `Configuration.make_sigmoid`, from which is should be picked up everywhere else

-  the problem reported by @lschwetlick is easily reproducible. The following snippet fails for `width_alpha > 0.05` 

```
import numpy as np
import psignifit

data = np.array([[0.0010, 45.0000, 90.0000],
                 [0.0015, 50.0000, 90.0000],
                 [0.0020, 44.0000, 90.0000],
                 [0.0025, 44.0000, 90.0000],
                 [0.0030, 52.0000, 90.0000],
                 [0.0035, 53.0000, 90.0000],
                 [0.0040, 62.0000, 90.0000],
                 [0.0045, 64.0000, 90.0000],
                 [0.0050, 76.0000, 90.0000],
                 [0.0060, 79.0000, 90.0000],
                 [0.0070, 88.0000, 90.0000],
                 [0.0080, 90.0000, 90.0000],
                 [0.0100, 90.0000, 90.0000]])

res = psignifit.psignifit(data, width_alpha=0.1)
```

- the exception is coming from _check_prior, which controls that probabilities are never 0
```
File ~/o/psignifit/psignifit/_priors.py:107, in _check_prior(priors, name, values)
    105     assert np.all(np.isfinite(result)), f"Prior '{name}' returns non-finite values."
    106     assert np.all(result >= 0), f"Prior '{name}' returns negative values."
--> 107     assert np.all(result != 0), f"Prior '{name}' returns zeros."
[...]
AssertionError: Prior 'width' returns zeros.
```

- there is nothing wrong in principle with the prior being 0 somewhere, in fact the prior for width is defined so that "the range of the stimulus levels with cosine fall offs to 0 at the minimal difference of two stimulus levels and at 3 times the range of the tested stimulus levels. Also, whether or not the prior is zero depends on the range on which the prior is calculated, which is different between `_priors._check_prior` and  when the posterior is computed in `_parameter.parameter_bounds`

- I think the check is there because in `_posterior.log_posterior` we compute the logarithm of the prior directly, which gives `-inf`  when the prior is 0
```
    # add priors on the corresponding axis
    p += np.log(priors['threshold'](thres))
[...]

PR #205, opened 2025-01-01
Fix alpha_width config

pberkes:
Partial fix for #204 , I implemented my recommendations for fixing the `alpha_width` configuration


otizonaizit (approved):
LGTM, thanks @pberkes for the work!

PR #208, opened 2025-01-04
Allow setting `thresh_PC` to something other than 0.5

### LINKED ISSUE #204: Options `thresh_PC` and `width_alpha` are still not working properly

pberkes:
Fix #204 

- turned the Exception raised when `thresh_PC != 0.5` into a meaningful warning
- fixed the position of the CI over threshold in `plot_psychometric_function` then `thresh_PC != 0.5`

### DIFF
--- a/docs/examples/advanced-options.md
+++ b/docs/examples/advanced-options.md
@@ -104,13 +104,9 @@ res = ps.psignifit(data, **options)
 
 ### Threshold definition (`thresh_PC`)
 
-```{warning}
-This option is only implemented in the MATLAB version of psignifit.
-For the python version the default value cannot be changed. 
-Adding this functionality is still work in progress.
-```
-
-This option sets the proportion correct to correspond to the threshold on the *unscaled* sigmoid. Possible values are in the range from 0 to 1, default is 0.5. The default corresponds to 75\% in a 2AFC task (midway between the guess rate of 50 % and ceiling performance 100%).
+This option sets the proportion correct to correspond to the threshold on the *unscaled* sigmoid. Possible values 
+are in the range from 0 to 1, default is 0.5. The default corresponds to 75\% in a 2AFC task (midway between the 
+guess rate of 50 % and ceiling performance 100%).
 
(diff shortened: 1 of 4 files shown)


otizonaizit:
Looks good and addresses my concerns mentioned in #204 .

Regarding the text of the warning: we are telling users that they must be aware of the fact that the prior over psychometric functions will be shifted. OK, but in addition to be aware of it, can the user do anything to address this? I mean, imagine I am a naive user and I want to set `thresh_PC` to `0.75`. I get this warning and I ask my self: 

> I don't know much about priors or posteriors, I use psignifit because it says that it is for "painfree estimation of psychometric functions", so, what can I do now that I am aware of this thing that I am not sure what it means?

pberkes:
@otizonaizit the message tells them to define a custom prior over the threshold... that's not going to be pain-free. 

I _think_ the default prior is what a user wants, in most cases: you are interested in estimating the stimulus level at which the percent correct crosses 75%, and your prior is that that point is somewhere within the stimulus range, because you set up the experiment in a meaningful way.

If that's the case, then we could have a comment in the documentation, and no warning. The warning would only confuse the user who did everything right.

pberkes:
@otizonaizit  how about this alternative message: 
"""
The thresh_PC parameter is set to {thresh_PC}, not the default 0.5. Be aware that the default prior over threshold assumes that the experimental stimulus levels cover the range where the threshold likely falls. If this doesn't match your setup, you'll need a custom prior. See the documentation for guidance.
"""

otizonaizit:
> @otizonaizit how about this alternative message: """ The thresh_PC parameter is set to {thresh_PC}, not the default 0.5. Be aware that the default prior over threshold assumes that the experimental stimulus range covers the range where the threshold likely falls. If this doesn't match your setup, you'll need a custom prior. See the documentation for guidance. """

Ohh yes!!! This is so much better!

guillermoaguilar (approved):

guillermoaguilar:
LGTM! 
I like the new warning.

PR #211, opened 2025-01-04
Reformulate docs

otizonaizit:
In this PR I plan to go over the whole documentation and fix simple inconsistencies, for example using *psignifit* when we talk about the library, and `psignifit` when we talk about the function in the module `psignifit`.

I also plan to remove the references to `python-psignifit`: this is only the name of the GitHub repo, but the package and everything else is called `psignifit`.

I also plan to remove the references to *psignifit 4*. The *4* has only an historical value and it's confusing if you are coming to the library nowadays. I adjusted the entry page to make it clear that this is an implementation of the Vision Research paper, so that it is clear what this is about. The fact that there were other *psigifit*'s in the past does not really matter, I think.

I also plan to remove the use of the very unpythonic:

```python
options = { 'something': something_else }

res = ps.psignifit(data, **options)
```

This patter only makes sense if the list of options you are passing is really big, which is hardly the case in most of our docs. I am changing it to:

```python
res = ps.psignifit(data, something=something_else)
```

Also, I have made every example runnable. In the process I have discovered several examples that were not run, but only presented as code in the text. Several of these examples would fail if run. These should all be fixed now.


otizonaizit:
Note that I have removed the page which was showing the comparison with the MATLAB implementation. Look at the commit message for a rationale:
005de7fc1b208af04db829b92b2582a2ecd9db7a

otizonaizit:
I still have to go through the quick how-tos and the extras. The getting started and the usage guides chapters are done

otizonaizit:
There are a lot of changes in the docs, as I explained above, most is only cosmetics. It is probably easier to just look at the built doc.

otizonaizit:
Also, this PR must be merged after #212 :-)

docs/how_to/How-to-Get-Standard-Parameters.md
```diff
@@ -13,22 +13,39 @@ kernelspec:
...
 
+res = ps.psignifit(data, sigmoid='norm', experiment_type='yes/no')
 ```
-expected_loc = threshold
+
+In this example we have a Gaussian sigmoid with `alpha=0.05`, `PC=0.5` fitted. The parameters we would get out, threshold and width, correspond to the standard parameters `loc` (mean) and `scale` (standard deviation) in the following way:
+
+```{code-cell} ipython3
+# the mean (loc):
+print(f'loc = {res.parameter_estimate['threshold']}')
```
pberkes, line 39: this breaks the docs compilation, because it's not valid code: it has nested single quotes. I'm not sure why it's not caught up by the CI, something else to investigate

docs/how_to/How-to-Get-Standard-Parameters.md
```diff
@@ -13,22 +13,39 @@ kernelspec:
...
+
+# the standard deviation (scale):
 # 1.644853626951472 is the normal PPF at alpha=0.95
-expected_scale = width / (2 * 1.644853626951472)
+print(f'scale = {res.parameter_estimate['width'] / (2 * 1.644853626951472)}')
```
pberkes, line 43: this breaks the docs compilation, because it's not valid code: it has nested single quotes

docs/how_to/How-to-Get-Standard-Parameters.md
otizonaizit: it doesn't break it for me, I think it is Python version dependent. Anyway, I'll fix it, the nested quotes are not necessary

docs/examples/map_vs_mean.md
```diff
@@ -115,16 +104,21 @@ Let's have a look at the results, starting from the posterior distribution over
 psp.plot_bayes(result);
 ```
 
-As you can see, the posterior is quite broad and asymmetrical, and we expect the mean and MAP estimates to be slightly different. We can ask `Result` for both estimate.
+As you can see, the posterior is quite broad and asymmetrical, and we expect the mean and MAP estimates to be slightly different. We can ask `result` for both estimate.
```
guillermoaguilar, line 107: for both estimate**S**

docs/how_to/How-to-Change-the-Threshold-Percent-Correct.md
```diff
@@ -17,13 +17,28 @@ This option sets the proportion correct to correspond to the threshold on the *u
 in the range from 0 to 1, default is 0.5. The default corresponds to 75\% in a 2AFC task (midway between the guess 
 rate of 50 % and ceiling performance 100%).
```
guillermoaguilar, line 19: add somewhere in the first paragraph that the option is called `thresh_PC`. Otherwise it's not so easy to see what it's changed in the arguments passed

docs/how_to/How-to-Get-Standard-Parameters.md
```diff
@@ -13,22 +13,39 @@ kernelspec:
...
+                 [0.0040, 62.0000, 90.0000], [0.0045, 64.0000, 90.0000],
+                 [0.0050, 76.0000, 90.0000], [0.0060, 79.0000, 90.0000],
+                 [0.0070, 88.0000, 90.0000], [0.0080, 90.0000, 90.0000],
+                 [0.0100, 90.0000, 90.0000]])
 
+res = ps.psignifit(data, sigmoid='norm', experiment_type='yes/no')
 ```
-expected_loc = threshold
+
+In this example we have a Gaussian sigmoid with `alpha=0.05`, `PC=0.5` fitted. The parameters we would get out, threshold and width, correspond to the standard parameters `loc` (mean) and `scale` (standard deviation) in the following way:
```
guillermoaguilar, line 35: I think here it's meant `thresh_PC=0.5` and `width_alpha=0.05`,
no?

docs/how_to/How-to-Get-Standard-Parameters.md
pberkes: I was using `Python 3.10.9`, is there a Python version where nested ingle quotes work? 😱  oh yes version 3.12! why!

pberkes: https://docs.python.org/3.12/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings

guillermoaguilar:
I went through all the changes, and made some minor comments, thank you @otizonaizit ! 

@lschwetlick:
There's one more thing I noticed in the parameter recovery demos.... content-wise. One shouldn't use "hits", it is 'count correct" or "number correct'.   The format for psignifit is: [stimulus value, # correct, # total].

This distinction is maybe neat-picky but "hits" is a concept from signal detection theory, and it is specific to it and not general. In the code it needs to be changed to "ncorrect".

otizonaizit:
I'll address the comments and fix the "hits" vs "count correct" thing. It may take a couple of days though.

lschwetlick:
Thanks for pointing it out @guillermoaguilar, i didnt think about the hits/count correct thing at all!

docs/examples/advanced-options.md
```diff
@@ -111,11 +103,16 @@ guess rate of 50 % and ceiling performance 100%).
...
 
 Note that this corresponds to a 95 \% in a 2AFC task.
 
+Note also that you will get a warning, because the default prior assumes that the experimental
+stimulus range covers the range where the threshold likely falls. If this doesn't match your
+setup, you'll need a custom prior. See the [priors demo](priors) to learn how to do it.
+
+
 
 ### Width definition (`width_alpha`)
```
lschwetlick, line 117: I think we should just have ##Threshold be one heading and ##Width Definition another, on the same level.

docs/examples/advanced-options.md
```diff
@@ -183,42 +182,5 @@ This for example would exclude fewer values and more conservative
...
-
-```{code-cell} ipython3
-options['move_bounds'] = False
-```
-
-
-## Other options
-### Pooling
-
-These options set how your data is pooled into blocks.
```
lschwetlick, line 206: *blocs

lschwetlick: throughout it should apparently be bloc and not block:
https://www.englishgrammar.org/bloc-vs-block/

docs/index.md
```diff
@@ -1,54 +1,37 @@
...
-or contact us: <heiko.schuett@uni-tuebingen.de> or
-<felix.wichmann@uni-tuebingen.de>
-
 ## Acknowledgements
 
-First, we would like to thank Sophie Laturnus and Ole Fortmann who
-programmed this python clone based on our MATLAB m-files.
+The first implementation, as Python-clone of the original MATLAB library, was done by Sophie Laturnus and Ole Fortmann. 
+
+The current completely rewritten implementation is the result of the work of several [contributors](https://github.com/wichmann-lab/python-psignifit/blob/main/CONTRIBUTORS).
```
lschwetlick, line 32: should we maybe name the sprint participants here explicitly?

lschwetlick (commented):
It looks good to me, thanks @otizonaizit

docs/examples/advanced-options.md
otizonaizit: Hmm, I'm not sure: bloc seems to refer to a group of people (or nations, or parties, so ... human things) and block to other things. In psychophysics for example you have block vs interleaved trial administration in experiments: 
https://pmc.ncbi.nlm.nih.gov/articles/PMC5319747/

guillermoaguilar: Yes, in psychophysics it's always "block", with K, I've never seen it without K.

lschwetlick: ok, then it should be block throughout though. Right now there is both. 

bloc makes more sense semantically, but i agree ive seen the ck more often.

otizonaizit:
Do not merge this PR yet. I want to rebase it on top of main once PR #212 has been discussed and either merged or refused.

Issue #64, opened 2021-12-01
Function to obtain the Percent Correct values for different stimulus levels?


emilaz:
I was wondering if we could have the ability to get the actual function out of the fitting process, meaning some way to get the percent correct values from a stimulus level? I haven't seen that anywhere in the docs so far.

dekuenstle:
I'm sorry for the late response. You are right, this is not straightforward at the moment. We are working on a refactored version, where this should become easier. 

You can construct the sigmoid handle to calculate your percentage correct with the fitted values. Something like:
```
import psignifit 

result = psignifit.psignifit(...)
sigmoid = psignifit.getSigmoidHandle.getSigmoidHandle(result['options'])
pc_at_x = sigmoid(x, result['Fit'][0],  result['Fit'][1])
``` 
Does this help so far?

pberkes:
This issue is still very much valid. At the moment, it involves multiple steps: creating a sigmoid from the Result object, then calling it using the estimated parameters.

Since we have "shortcut" methods on the Result object for threshold, slope, and parameters estimate, I suggest adding a `Result.percent_correct(stimulus_level, estimate_type)` method

PR #212, opened 2025-01-05
cast estimated parameters from Numpy scalars  to Python floats

otizonaizit:
Do not return Numpy scalars, i.e. 0-D Numpy arrays. Use Python floats instead.
This does not have any numerical relevance, but makes the returned values easier to read, especially during interactive work and in the documentation. I am using this PR already in #211 .

I have solved this by introducing a new utility `cast_np_scalar`, which can be called unconditionally on anything. If the input is a numpy scalar it will be converted to a Python scalar (typically, float or int), otherwise it is a no-op. This is necessary because we have functions and methods both in `Result` and `Sigmoid` which work both with Numpy arrays and scalars. We don't want to touch the arrays.

All this is needed because Numpy 2 implemented [NEP 51](https://numpy.org/neps/nep-0051-scalar-representation.html) and numpy scalars have a repr of the form `np.float64(1.2)`, i.e. they do not look like Python scalars anymore. Numpy returns a scalar as a result of operations on arrays where the dimensionality reduction due to the operation would lead to a 0-D array, for example `array.sum()`.

### DIFF
--- a/psignifit/_posterior.py
+++ b/psignifit/_posterior.py
@@ -250,7 +250,10 @@ def maximize_posterior(data, param_init: Dict[str, float], param_fixed: Dict[str
     init_values = [value for name, value in sorted(param_init.items())]
     optimized_values = optimize.fmin(objective, init_values, disp=False)
     # "sorted" is required so that the parameters are matched to the axes in  alphabetical order
-    optimized_param = dict(zip(sorted(param_init.keys()), optimized_values))
+    optimized_param = dict(zip(sorted(param_init.keys()), optimized_values.tolist()))
+    # convert fixed parameters to Python floats (they are numpy scalars)
+    for key, value in param_fixed.items():
+        param_fixed[key] = float(value)
     return {**param_fixed, **optimized_param}
 
 
--- a/psignifit/_posterior.py
+++ b/psignifit/_posterior.py
@@ -5,8 +5,8 @@ import numpy as np
 import scipy.special as sp
 from scipy import optimize
 
-from ._utils import fp_error_handler, PsignifitException, cast_np_scalar
-from ._typing import Prior, ParameterGrid
+from ._utils import PsignifitException, cast_np_scalar, fp_error_handler
+from ._typing import ParameterGrid, Prior
 from .sigmoids import Sigmoid
 
 # accomodate numpy versions < 2
--- a/psignifit/_posterior.py
+++ b/psignifit/_posterior.py
@@ -5,7 +5,7 @@ import numpy as np
 import scipy.special as sp
 from scipy import optimize
 
-from ._utils import fp_error_handler, PsignifitException
+from ._utils import fp_error_handler, PsignifitException, cast_np_scalar
 from ._typing import Prior, ParameterGrid
 from .sigmoids import Sigmoid
 
@@ -248,10 +248,11 @@ def maximize_posterior(data, param_init: Dict[str, float], param_fixed: Dict[str
 
     # "sorted" is required so that the parameters are matched to the axes in  alphabetical order
     init_values = [value for name, value in sorted(param_init.items())]
-    optimized_values = optimize.fmin(objective, init_values, disp=False)
+    optimized_values = [cast_np_scalar(value) for value in optimize.fmin(objective, init_values, disp=False)]
     # "sorted" is required so that the parameters are matched to the axes in  alphabetical order
     optimized_param = dict(zip(sorted(param_init.keys()), optimized_values))
-    return {**param_fixed, **optimized_param}
+    out_param_fixed = { key: cast_np_scalar(value) for key, value in param_fixed.items()}
+    return {**out_param_fixed, **optimized_param}
 
 
 def marginalize_posterior(parameter_grid: ParameterGrid, posterior_mass: np.ndarray) -> Dict[str, np.ndarray]:
--- a/psignifit/_posterior.py
+++ b/psignifit/_posterior.py
@@ -248,11 +248,12 @@ def maximize_posterior(data, param_init: Dict[str, float], param_fixed: Dict[str
 
     # "sorted" is required so that the parameters are matched to the axes in  alphabetical order
     init_values = [value for name, value in sorted(param_init.items())]
-    optimized_values = [cast_np_scalar(value) for value in optimize.fmin(objective, init_values, disp=False)]
+    optimized_values = optimize.fmin(objective, init_values, disp=False)
     # "sorted" is required so that the parameters are matched to the axes in  alphabetical order
     optimized_param = dict(zip(sorted(param_init.keys()), optimized_values))
-    out_param_fixed = { key: cast_np_scalar(value) for key, value in param_fixed.items()}
-    return {**out_param_fixed, **optimized_param}
+    cast_optimized_param = { key: cast_np_scalar(value) for key, value in optimized_param.items() }
+    cast_param_fixed = { key: cast_np_scalar(value) for key, value in param_fixed.items() }
+    return {**cast_param_fixed, **cast_optimized_param}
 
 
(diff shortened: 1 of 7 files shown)


pberkes:
@otizonaizit  I think the changes are good... but it would be really easy to add two small tests checking `isinstance(x, float)` to avoid future regressions

otizonaizit:
> @otizonaizit I think the changes are good... but it would be really easy to add two small tests checking `isinstance(x, float)` to avoid future regressions

working on it, it is harder than I thought to reproduce the old behavior in a test

otizonaizit:
> > @otizonaizit I think the changes are good... but it would be really easy to add two small tests checking `isinstance(x, float)` to avoid future regressions
> 
> working on it, it is harder than I thought to reproduce the old behavior in a test

ahah, that is why it was hard to reproduce:
```python
>>> import numpy as np
>>> print(np.__version__)
2.1.3
>>> x = np.float64(0.00012925277952539202)
>>> print(x)
0.00012925277952539202
>>> repr(x)
'np.float64(0.00012925277952539202)'
>>> isinstance(x, float)
True
```

vs.

```python
>>> import numpy as np
>>> print(np.__version__)
1.26.4
>>> x = np.float64(0.00012925277952539202)
>>> print(x)
0.00012925277952539202
>>> repr(x)
'0.00012925277952539202'
>>> isinstance(x, float)
True
```

time to investigate...


I need to investigate this

otizonaizit:
This is the corresponding NEP:
https://numpy.org/neps/nep-0051-scalar-representation.html

otizonaizit:
So, `np.float64` objects are a subclass of Python floats:

```python
>>> import numpy as np
>>> print(np.__version__)
2.1.3
>>> x = np.float64(10.)
>>> x.__class__.__mro__
(<class 'numpy.float64'>, <class 'numpy.floating'>, <class 'numpy.inexact'>, <class 'numpy.number'>, <class 'numpy.generic'>, <class 'float'>, <class 'object'>)
>>> isinstance(x, float)
True
>>> type(x) is float
False
>>> type(float(x)) is float
True
```

so I am not sure there is a way to test that we indeed return Python scalars using `isinstance`. I will implement the tests using `type`, but I am not sure if being so strict will cause other problems down the road...

pberkes:
maybe `isinstance(x, float) and not isinstance(x, np.floating)`?

otizonaizit:
> maybe `isinstance(x, float) and not isinstance(x, np.floating)`?

It's a can of worms. But I'm working on it. I'm coming up with a general solution. Hopefully with as few lines of code as possible.

otizonaizit:
@pberkes : this is my solution. What do you think? I tested with both Numpy < 2 and >= 2 and it works unchanged in both cases, so I think it's safe.

otizonaizit:
> @pberkes : this is my solution. What do you think? I tested with both Numpy < 2 and >= 2 and it works unchanged in both cases, so I think it's safe.

I have updated the PR text for the record.

tests/test_numpy_scalars.py
```diff
@@ -0,0 +1,78 @@
...
+from .fixtures import input_data
+from psignifit.sigmoids import ALL_SIGMOID_CLASSES
+
+@pytest.mark.parametrize('num', (1, 1., np.int64(1), np.float64(1.)))
+def test_cast_np_scalar_numbers(num):
+    cast = cast_np_scalar(num)
+    assert not isinstance(cast, np.number)
+    assert isinstance(cast, numbers.Number)
+
+def test_cast_np_scalar_ndarray_noop():
```
pberkes, line 17: please 2 spaces between functions everywhere (PEP8)

psignifit/_posterior.py
```diff
@@ -5,7 +5,7 @@
 import scipy.special as sp
 from scipy import optimize
 
-from ._utils import fp_error_handler, PsignifitException
+from ._utils import fp_error_handler, PsignifitException, cast_np_scalar
```
pberkes, line 8: alphabetical order please

psignifit/_posterior.py
```diff
@@ -248,10 +248,11 @@ def objective(x):
 
     # "sorted" is required so that the parameters are matched to the axes in  alphabetical order
     init_values = [value for name, value in sorted(param_init.items())]
-    optimized_values = optimize.fmin(objective, init_values, disp=False)
+    optimized_values = [cast_np_scalar(value) for value in optimize.fmin(objective, init_values, disp=False)]
```
pberkes, line 251: This is correct but makes the line less readable, the line is important and its main intent is finding the MAP point estimate.
How about leaving it as it is, and at the end of the function declaring a `out_optimized_values` that does the casting, as in `out_param_fixed`? and they could be calles `cast_optimized_values` and `case_param_fixed`

psignifit/psignifit.py
```diff
@@ -13,7 +13,7 @@
 from ._priors import setup_priors
 from ._result import Result
 from ._typing import ParameterBounds, Prior
-from ._utils import (PsignifitException, check_data)
+from ._utils import PsignifitException, check_data, cast_np_scalar
```
pberkes, line 16: alphabetical order?

pberkes (commented):
Thank you, @otizonaizit , the code looks good. I only have nitpicking style comments that you can ignore if you want (except for the 2 spaces between functions :-P )

I had the time to look into NEP 51... In NEP 51, they only changed the `repr` of numpy scalars. It does seem a bit overkill to convert all the numpy scalars to actual Python floats, only to change the printout in interactive python sessions.

The same effect can be achieved with `np.set_printoptions(legacy='1.25')` , so maybe that can be used while building the documentations?

otizonaizit:
I can take care of the nitpicks.

Regarding the option of just changing the printing style in the docs... I am afraid this will be even more confusing for users. If they run the stuff locally they will see something different than in the docs.

I agree that this PR is a bit overkill, but I am worried that our users, which will be mostly people without a background in programming, will be confused by the `np.float64(1.)` outputs. We could just set the NumPy printing options directly in psignifit, but this seems to be a global option, so we would mess up with the users' global environment, which is a no-go...

All in all I think the changes are not so big to make it a problem to maintain the code down the line.

tests/test_numpy_scalars.py
otizonaizit: fixed in 5f66c62036dab3b87661e8827d2715407021440a

psignifit/_posterior.py
otizonaizit: fixed in 326c7caa01205a68d86463c5bc55402a90b1999a

psignifit/psignifit.py
otizonaizit: fixed in 326c7caa01205a68d86463c5bc55402a90b1999a

otizonaizit:
> @otizonaizit I think it's fine because we're only using `float64` everywhere. It's ok with me even if I feel somewhat dirty about it

yeah, but good luck explaining what `float64` means and why it is wrapping a number like in a function call... for beginners something like `np.float64(1)` is a function call, and they expect it to return something. I agree also that it only happens with `repr`, which unfortunately is what jupyter and the Python prompt shows you by default if you don't use `print`...

otizonaizit:
but we have to come to a decision, because I am working on another PR to fix #64 , which can or can not use this PR...

pberkes:
What I meant is that if we used other float precisions, then the cast would have been more disruptive. AFAIC, this PR can be merged

otizonaizit:
> What I meant is that if we used other float precisions

I see... yes, even if I think that the casts are only done for when the functions return single values, so imagine you are building a pipeline using `float32`, you'll already have to make sure to cast your scalars, including the literals you may use everywhere in the code. If you pass a `float32`-array into `Sigmoid.__call__` I am not sure what happens right now, but it could well be that it gets casted already to `float64` just because we multiply by `gamma`.
You think it is worth opening an issue so that we don't forget and can discuss about it sometime later?

pberkes:
@otizonaizit  no, no, it's fine

PR #215, opened 2025-01-08
build docs using oldest supported Python version and add Python 3.13 to CI tests and release workflows

### LINKED ISSUE #213: In CI, compile docs using the oldest supported Python version
At the moment, we compile the docs using Python 3.13, but in the CI tests we support 3.10, 3.11, and 3.12 .

The docs need to be compiled using the oldest supported Python version, to make sure the examples can be executed by users with older versions. 

This caused a problem, for instance, with nested f-string single quotes in #211

otizonaizit:
Fixes #213

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.13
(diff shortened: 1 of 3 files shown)


pberkes:
The docs failed. Judging from the error, I think you need to put quoted around "3.10", or it's interpreted as a float and becomes "3.1"

PR #224, opened 2025-01-10
Add to the matrix Python==3.12, numpy==1.26.4

### LINKED ISSUE #217: Add last NumPy 1.X release to CI matrix
After the release of Numpy [2.0](https://numpy.org/doc/2.1/release/2.0.0-notes.html), backward compatibility with the [last version](https://numpy.org/doc/2.0/release/1.26.4-notes.html) in the NumPy 1.X series is not given anymore and we have already code to cope with some deleted names, for example here:
https://github.com/wichmann-lab/python-psignifit/blob/0fec2d93a5c3193429dc2885d3ea6edf9205d8f8/psignifit/_posterior.py#L12
and here:
https://github.com/wichmann-lab/python-psignifit/blob/0fec2d93a5c3193429dc2885d3ea6edf9205d8f8/psignifit/_priors.py#L11

The NumPy 1.X versions are still very much in use in the wild, for example Ubuntu/Debian based Linux distros and many scientific Python packages that haven't been updated to support the NumPy 2.X series. I think we should support it until it is feasible to do so without too much clutter in our code.

I propose to add an item in the test matrix in the CI to cover for this.

All in all, I think we should test:

  - Python `3.12` + NumPy `1.26.4` (`3.12` is the highest Python version supported by NumPy `1.26.4`)
  - Python `3.10, 3.11, 3.12, 3.13` + Numpy latest (`2.2.1` as of now, but to get the latest we can rely on `pip` picking the last version by default)

pberkes:
Fix #217 

Fingers crossed

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -17,6 +17,11 @@ jobs:
       matrix:
         os: [Ubuntu]
         python_version: ['3.10', '3.11', '3.12', '3.13']
+        numpy_version: ['2.*']
+        include:
+          - python_version: '3.12'
+            numpy_version: '1.26.4'
+            os: 'Ubuntu'
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


otizonaizit:
Great! Where did you learn the trick with specifying the version with `*`? This is cool, especially because we are already forward compatible with the breaking release of NumPy 3.0 scheduled for the next century :-)))

PR #225, opened 2025-01-10
Fixed parms tests

### LINKED ISSUE #221: Enhance test for fixed parameters
Fixed parameters can at the moment be set to non-sensical values, for example `width=0`, which results in some random exceptions along the line due to infinities or NaNs appearing everywhere. We need to catch this early on, ideally right in the `Configuration` object and fail with a proper error message.

otizonaizit:
Error and warning messages and sensible bounds for the parameters were discussed with @FelixWichmann .

Fixes #221

### DIFF
--- a/psignifit/_configuration.py
+++ b/psignifit/_configuration.py
@@ -155,6 +155,10 @@ class Configuration:
                     # 0 <= gamma < 1 and 0 <= lambda < 1
                     raise PsignifitException(f'{prefix} 0 <= {parm_name} < 1 {suffix}')
                 # WARNINGS
+                if parm_name == 'lambda' and parm_value > 0.2:
+                    # lambda > 0.2 is unusual
+                    warnings.warn(f"You have fixed the lapse rate lambda to an unusually high value ({parm_value}). "
+                                  f"In typical psychophysical experiments lambda is rarely above 0.2 for human observers.")
 
     def check_experiment_type_matches_fixed_parameters(self, fixed_params, experiment_type):
         if experiment_type == ExperimentType.N_AFC.value:
--- a/psignifit/_configuration.py
+++ b/psignifit/_configuration.py
@@ -139,7 +139,7 @@ class Configuration:
                 # 0 <= gamma + lambda < 1
                 if not (0 <= gamma+lambda_ < 1):
                     raise PsignifitException(f'For gamma and lambda the condition'
-                                             f' 0 <= gamma + lambda < 1 mut always apply:'
+                                             f' 0 <= gamma + lambda < 1 must always apply:'
                                              f' got gamma={gamma}, lambda={lambda_} instead!')
             for parm_name, parm_value in value.items():
                 # ERRORS
--- a/psignifit/_configuration.py
+++ b/psignifit/_configuration.py
@@ -165,7 +165,7 @@ class Configuration:
             if fixed_params is not None and 'gamma' in fixed_params:
                 warnings.warn(
                     f'The parameter gamma was fixed to {fixed_params["gamma"]}. In {ExperimentType.N_AFC.value} experiments gamma is automatically fixed to 1/n. Ignoring fixed gamma.')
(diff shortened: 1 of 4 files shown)


psignifit/_configuration.py
```diff
@@ -131,16 +131,41 @@ def check_fixed_parameters(self, value):
             vkeys = _PARAMETERS
             if vkeys < set(value.keys()):
                 raise PsignifitException(
-                    f'Option fixed_paramters keys must be in {vkeys}. Given {list(value.keys())}!'
+                    f'Option fixed_parameters keys must be in {vkeys}. Given {list(value.keys())}!'
                 )
-
+            # Check that the fixed parameters have been set to meaningful values
+            if (gamma := value.get('gamma')) and (lambda_ := value.get('lambda')):
```
pberkes, line 137: in these `if`s, I think you want to check if the key has been defined. You need to use `if 'gamma' in value` instead.

For example, if `value = {'gamma': 0.0}`, `value.get('gamma')` is 0.0 and so evaluated as False.

In this snippet, the print statement will not be executed:
```
value = {'gamma': 0.0}
if value.get('gamma'): print('alright')
```

psignifit/_configuration.py
```diff
@@ -131,16 +131,41 @@ def check_fixed_parameters(self, value):
...
+                # this condition can only happen if both lambda and gamma are fixed
+                # 0 <= gamma + lambda < 1
+                if not (0 <= gamma+lambda_ < 1):
+                    raise PsignifitException(f'For gamma and lambda the condition'
+                                             f' 0 <= gamma + lambda < 1 mut always apply:'
```
pberkes, line 142: "mut" -> "must"

psignifit/_configuration.py
pberkes: Case where it would fail: gamma=0, lambda=1 would not raise an exception

pberkes (changes requested):
Thank you @otizonaizit !

I requested a couple of changes

psignifit/_configuration.py
otizonaizit: I added two tests in 5fc7e8410c6d53945795fdca0f3ed07786189c63 which show that the correct `Exception` is thrown.

It is true that the code in the if-block is not executed if one of the two parameters is exactly zero, but in that case the code below will make sure that the other, non-zero, parameters is strictly smaller than 1, so at the end everything works as expected. I could change the comment in the if to make it more clear?

Or make the code more verbose as you propose... I don't know. Somehow I think it's nicer that if the input is `gamma=0, lambda=2` the error complains about `lambda` alone and not about the sum of `gamma + lambda`, but it's a matter of taste...

What do you think?

pberkes: Yes, the code still works ... as a side effect

My issue is in the meaning of the if test. `if value.get('gamma')` means "if the value of value['gamma'] evaluates to False (and if the key 'gamma' is missing, that's the case)". `if 'gamma' in value` means "if the key 'gamma' is defined).

Then if you want to have an error about lambda alone in case `gamma=0, lambda=2`, that can also be arranged by checking the individual keys first. It is a separate concern

It's a matter of having the code more transparent, please leave it like it if you think I'm nitpicking. The code works.

otizonaizit: fixed in 7280b470a87052d2562565aa4589bddd824dbb9c

pberkes (approved):
Looks good to me

psignifit/_configuration.py
```diff
@@ -160,6 +153,16 @@ def check_fixed_parameters(self, value):
                     warnings.warn(f"You have fixed the lapse rate lambda to an unusually high value ({parm_value}). "
                                   f"In typical psychophysical experiments lambda is rarely above 0.2 for human observers.")
 
+            if 'gamma' in value and 'lambda' in value:
+                gamma, lambda_ = value['gamma'], value['lambda']
```
pberkes, line 157: :=)

PR #233, opened 2025-01-26
Add couple of more datasets to test against

### LINKED ISSUE #61: Tests to compare matlab and python versions
1. Specify multiple configurations & data 
2. Run these matlab psignifit and store the result
3. Compare the fitted parameters in python psignifit with the stored matlab fit.

guillermoaguilar:
These datasets were simulated and are included in the matlab version of psignifit. The tests check against the values obtained by matlab, with a "sensible" tolerance (1% relative tolerance, 0.1% absolute tolerance). With sensible I mean a tolerance that is in the range of what a user cares. 

Because there are already extensive tests that check for parameter recovery, I include here only some of the datasets:
- 2AFC and gaussian
- yes/no and gaussian
- 2AFC and logistic
- yes/no and logistic

The point is to have a rough check that the matlab and the python version return the same values, within a "sensible" range.

This PR closes #61

### DIFF
--- a/tests/test_psignifit.py
+++ b/tests/test_psignifit.py
@@ -18,11 +18,11 @@ def test_psignifit_debug_info(input_data):
 
 
 def test_psignifit_fixed_fit(input_data):
-    parm_matlab= {'threshold': 0.0046448472488663396,
-                  'width': 0.004683837353547434,
-                  'lambda': 1.0676339572811912e-07,
-                  'gamma': 0.5,
-                  'eta': 0.00011599137494786461}
+    parm_matlab = {'threshold': 0.0046448472488663396,
+                   'width': 0.004683837353547434,
+                   'lambda': 1.0676339572811912e-07,
+                   'gamma': 0.5,
+                   'eta': 0.00011599137494786461}
 
     results = psignifit(input_data, sigmoid='norm', experiment_type='2AFC')
     parm = results.get_parameter_estimate(estimate_type='MAP')
--- a/tests/test_psignifit.py
+++ b/tests/test_psignifit.py
@@ -18,7 +18,7 @@ def test_psignifit_debug_info(input_data):
 
 
 def test_psignifit_fixed_fit(input_data):
-    parm_heiko = {'threshold': 0.0046448472488663396,
+    parm_matlab= {'threshold': 0.0046448472488663396,
                   'width': 0.004683837353547434,
                   'lambda': 1.0676339572811912e-07,
                   'gamma': 0.5,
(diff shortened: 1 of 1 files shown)


guillermoaguilar:
And this PR should make PR #66 obsolete

tests/test_psignifit.py
```diff
@@ -18,7 +18,7 @@ def test_psignifit_debug_info(input_data):
 
 
 def test_psignifit_fixed_fit(input_data):
-    parm_heiko = {'threshold': 0.0046448472488663396,
+    parm_matlab= {'threshold': 0.0046448472488663396,
```
pberkes, line 21: Missing space around '='

pberkes (approved):
Looks great, thank you!

tests/test_psignifit.py
otizonaizit: Regarding this kind of nitpick, I think it is mostly a matter of personal preference. Yes, the spacing on this line is not consistent, but it is a line in a test file and it is not exposed in documentation or to the casual source code explorer. In cases like this I think the nitpicker, in this case @pberkes , could make use of the power of just editing the PR himself. If the nitpicker doesn't think it's worth it to make th effort, then it is even less worth it for the author of the PR.
Again, this is just my opinion, but I think it makes the atmosphere in general nicer and friendlier. Also, it distributes the load of personal preferences on those who _have_ the preference ;-)

otizonaizit:
Great @guillermoaguilar : this one was the last roadblock to the 4.3 release, right? Congratulations!

otizonaizit (approved):

tests/test_psignifit.py
pberkes: ok I've done it. Personally, I think it's best for the author to do the editing because it avoids having to think about push/pull/merge on your own PR made on your own fork. it takes 10 seconds to fix this kind of PEP8 violations.

otizonaizit: sure, it takes 10 seconds for you or for the author. My point here is, if we use this PR as example:
  - @guillermoaguilar spends time writing the code, tests it and submits a PR, then goes on about his business
  - we review it and the only thing we find worth changing is a minor PEP8 violation in a test file

To me it seems obvious that while you do the review you can fix the minor thing and just merge the PR. This way @guillermoaguilar is not involved anymore: I don't know about his schedule, but it happens to me quite often to make a PR and then for the next two days not to have time to do anything more with it, so for me at least it would really help if you would edit and merge, so I don't have the feeling of being a roadblock.

In cases where there are substantial requests of changes in the code the picture is completely different. In that case I agree that the changes should be left to the author of the PR.

PR #239, opened 2025-06-15
fixes bug in x-range of posterior samples plot

### LINKED ISSUE #238: In posterior samples plot, stimulus range on the x-axis is not read from the data
but it is hardcoded between 0.001 and 0.01. So for the documentation example it works fine, but not for other datasets, e.g.

Image

guillermoaguilar:
fixes #238 

fixed_bug

### DIFF
--- a/psignifit/psigniplot.py
+++ b/psignifit/psigniplot.py
@@ -669,7 +669,7 @@ def plot_posterior_samples(
 
     # Plot the samples from the posterior
     sigmoid = result.configuration.make_sigmoid()
-    x = np.linspace(0.001, 0.01, num=1000)
+    x = np.linspace(result.data[:, 0].min(), result.data[:, 0].max(), num=1000)
     for idx in range(n_samples):
         y = sigmoid(
             x,


psignifit/psigniplot.py
```diff
@@ -678,7 +678,7 @@ def plot_posterior_samples(
             gamma=params_samples['gamma'][idx],
             lambd=params_samples['lambda'][idx],
         )
-        ax.plot(x, y, alpha=samples_alpha, color=samples_color)
+        ax.plot(x, y, alpha=samples_alpha, color=samples_color, zorder=-1)
```
pberkes, line 681: As far as I understand, a negative z-order would place the samples behind the plotting elements of the axis, which might make them disappear if the user sets a non-transparent background for the plot. I think a value of 0.5 might be better (e.g. https://matplotlib.org/stable/gallery/misc/zorder_demo.html#sphx-glr-gallery-misc-zorder-demo-py)

pberkes (approved):
LGTM, I recommend a change for the z-order value

PR #241, opened 2025-06-17
Adapts threshold method to return worse-case confidence intervals correctly

### LINKED ISSUE #191: Credible intervals incorrect for values below threshold?
# Context
Great work on psignifit. While reproducing some of my results, however, I unfortunately stumbled across a potential issue with the credible intervals. Below a code example (three versions) which shows that credible intervals for performance values below the threshold seem to be incorrect. 

# Minimal example
```
import numpy as np
import psignifit as ps
from psignifit import psigniplot
import matplotlib.pyplot as plt


plotVersion = 3  # select which of the proposed versions to plot (1-3)

# Example from documentation:
data = np.array([[0.0010, 45.0000, 90.0000], [0.0015, 50.0000, 90.0000],
                 [0.0020, 44.0000, 90.0000], [0.0025, 44.0000, 90.0000],
                 [0.0030, 52.0000, 90.0000], [0.0035, 53.0000, 90.0000],
                 [0.0040, 62.0000, 90.0000], [0.0045, 64.0000, 90.0000],
                 [0.0050, 76.0000, 90.0000], [0.0060, 79.0000, 90.0000],
                 [0.0070, 88.0000, 90.0000], [0.0080, 90.0000, 90.0000],
                 [0.0100, 90.0000, 90.0000]])

options = {'sigmoid': 'norm',
           'experiment_type' : '2AFC'}

res = ps.psignifit(data, **options)

plt.figure()
psigniplot.plot_psychometric_function(res)


# Plot credible intervals across the whole curve
if plotVersion == 1: # produces an error
    y = np.linspace(0.5, 1, 50)  # get credible intervals between 50-100% performance
    CIlow = []; CIhigh = []
    for i in y:
        threshold, CI = res.threshold(i)  # use scaled sigmoid
        CIlow.append(CI[0][0]); CIhigh.append(CI[0][1])
    plt.plot(CIlow, y, label="lower bound"); plt.plot(CIhigh, y, label="upper bound")

elif plotVersion == 2:
    y = np.linspace(0.02, .5, 50) # get credible intervals according to error message of version 1
    CIlow = []; CIhigh = []
    for i in y:
        threshold, CI = res.threshold(i) # use scaled sigmoid
        CIlow.append(CI[0][0]); CIhigh.append(CI[0][1])
    plt.plot(CIlow, y+0.5, label="lower bound"); plt.plot(CIhigh, y+0.5, label="upper bound")

elif plotVersion == 3:
    y = np.linspace(0., 1, 50)  # get credible intervals over whole range for unscaled sigmoid
    CIlow = []; CIhigh = []
    for i in y:
        threshold, CI = res.threshold(i, unscaled=True) # use unscaled sigmoid instead
        CIlow.append(CI[0][0]); CIhigh.append(CI[0][1])
    
    y = np.linspace(0.5, 1, 50) # redefine to 50-100%
    plt.plot(CIlow, y, label="lower bound"); plt.plot(CIhigh, y, label="upper bound")

plt.legend()
plt.show()
```

# Expected behavior
I would have expected that the psychometric function is always inbetween the credible intervals across the whole performance-range.

# Observed behavior
The credible intervals for performance values below threshold (75% in the example) dont include the fitted function values.

## Version 2 of code
image

## Version 3 of code
image

guillermoaguilar:
Partially fixes #191 

However, the function does not take a vectorized form for proportion_correct

And the tests need to be changed, the expected values were manually changed.

I need help with that, could you help @pberkes ?

### DIFF
--- a/psignifit/_result.py
+++ b/psignifit/_result.py
@@ -124,8 +124,7 @@ class Result:
             proportion_correct_unscaled = proportion_correct
         else:
             lambd, gamma = estimate['lambda'], estimate['gamma']
-            proportion_correct_unscaled = (proportion_correct - gamma)/(1- lambd - gamma)
-            
+            proportion_correct_unscaled = (proportion_correct - gamma)/(1- lambd - gamma)     
         new_threshold = sigmoid.inverse(proportion_correct, estimate['threshold'],
                                         estimate['width'], gamma, lambd)
         if not return_ci:
--- a/psignifit/_result.py
+++ b/psignifit/_result.py
@@ -121,8 +121,11 @@ class Result:
         estimate = self.get_parameter_estimate(estimate_type)
         if unscaled:  # set asymptotes to 0 for everything.
             lambd, gamma = 0, 0
+            proportion_correct_unscaled = proportion_correct
         else:
             lambd, gamma = estimate['lambda'], estimate['gamma']
+            proportion_correct_unscaled = (proportion_correct - gamma)/(1- lambd - gamma)
+            
         new_threshold = sigmoid.inverse(proportion_correct, estimate['threshold'],
                                         estimate['width'], gamma, lambd)
         if not return_ci:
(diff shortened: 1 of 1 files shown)


pberkes:
@guillermoaguilar I can fix the tests

psignifit/_result.py
```diff
@@ -138,9 +155,24 @@ def threshold(self, proportion_correct: np.ndarray, unscaled: bool = False, retu
...
+
+            ci_min[mask_above] = sigmoid.inverse(proportion_correct[mask_above],
+                                                 thres_ci[0], width_ci[0], gamma_ci[0], lambd_ci[1])
+            ci_max[mask_above] = sigmoid.inverse(proportion_correct[mask_above],
+                                                 thres_ci[1], width_ci[1], gamma_ci[1], lambd_ci[0])
+
+            ci_min[~mask_above] = sigmoid.inverse(proportion_correct[~mask_above],
+                                                 thres_ci[0], width_ci[1], gamma_ci[0], lambd_ci[1])
+            ci_max[~mask_above] = sigmoid.inverse(proportion_correct[~mask_above],
+                                                 thres_ci[1], width_ci[0], gamma_ci[1], lambd_ci[0])
```
pberkes, line 172: This implementation results in the threshold CI in the `test_threshold_value` to have a maximum larger than the minimum:

```
thr_ci['0.95'] = array([[0.792067], [0.580263]])
```

pberkes: I believe we need take the smallest `gamma` and `lambda` for the minimum, and the largest for the maximum

psignifit/_result.py
```diff
@@ -115,14 +117,29 @@ def threshold(self, proportion_correct: np.ndarray, unscaled: bool = False, retu
             (thresholds, ci): stimulus values along with confidence intervals
 
         """
+        if return_ci:
+            warnings.warn("""The confidence intervals computed by this method are only upper bounds. 
+                          To get a more accurate confidence interval at another level of proportion
```
pberkes, line 122: proposal: "To get a more accurate confidence interval at a level of proportion  correct other than `thresh_PC`"

The original sentence does not communicate what the reference "level of proportion correct is"

pberkes (commented):
I think there's a problem with the way the CI bounds for gamma and lambda are set, I'm pushing a proposal that fixes it in a minute

psignifit/_result.py
guillermoaguilar: yes agreed

psignifit/_result.py
guillermoaguilar: yep you're right, I just checked both cases and as it is now implemented returns the most conservative values.

PR #244, opened 2025-07-20
Add test to reproduce issue #243, and fix

### LINKED ISSUE #243: pool_blocks crashes on single-trial blocks
Minimal repro:
```
bla2 = np.array([
    [100, 50, 50],
    [101, 51, 51],
    [200, 60, 60],
    [201, 61, 61],
    [300, 70, 70],
    [301, 71, 71],
    [400, 80, 80]
])

# Crashes with:
# TypeError: cannot unpack non-iterable numpy.float64 object
pooled = ps.pool_blocks(bla2, max_tol=50, max_gap=50, max_length=10)
```
Two things:
- max_length refers to cumulative trial count (not number of rows), which isn’t super clear in the doc.
- If pooling results in just one trial (because one line already exceeds the max_length), the code crashes — probably needs to reshape the block before summing.

we should either error gracefully or reshape before.

pberkes:
Fix #243 

We need to make sure that there is at least one line in each block. The fix matches the condition in the Matlab source core
https://github.com/wichmann-lab/psignifit/blob/5f7241f6288c7dd3e1b34353118ca919ad28e51b/private/poolData.m#L32

### DIFF
--- a/psignifit/tools.py
+++ b/psignifit/tools.py
@@ -34,9 +34,11 @@ def pool_blocks(data: np.ndarray, max_tol=0, max_gap=np.inf, max_length=np.inf):
             block = []
             gap = 0
             for j in range(i, ndata):
-                if (cum_ntrials[j + 1] -
-                        cum_ntrials[i]) > max_length or gap > max_gap:
+                # j > i condition: make sure there is at least one data point per block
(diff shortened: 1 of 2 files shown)


guillermoaguilar (approved):
LGTM

